PR #84, opened 2026-04-25
Add fan speed slider support

wonderslug:
## Summary

- Extend the slider control in `button_grid.h` to route `fan.*` entities to `fan.turn_on` / `fan.turn_off` with a `percentage` parameter, instead of falling through to the light handler
- Subscribe to the `percentage` attribute (0-100) for state sync instead of brightness

## Test plan

- [ ] Slider controls a fan entity in Home Assistant and reflects current speed
- [ ] Slider still works correctly for light entities (no regression)
- [ ] Fan turns off when slider is set to 0

### DIFF
--- a/builds/guition-esp32-p4-jc1060p470.factory.yaml
+++ b/builds/guition-esp32-p4-jc1060p470.factory.yaml
@@ -17,11 +17,11 @@ packages:
 
 esphome:
   project:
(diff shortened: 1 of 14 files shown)


wonderslug:
Sorry, left dev changes in there, resubmitting

PR #87, opened 2026-04-25
Add Central European Latin glyph support

P1-Ro:
This pull request expands the extended Latin character set used by the UI to fully support Central European languages, specifically Slovak, Czech, Hungarian, and Polish.


jtenniswood:
Fab, many thanks

Issue #78, opened 2026-04-24
Change Buttons / Screen from HA


ToBeHH:
It would be great if Home Assistant could trigger changes on the screen.

So my scenario is that I have the ESP screen in my home office, and I would like to have different buttons depending on if I am working (e.g. switch everything off) or not. One way to do this could be that I define different screens in ESP control and then have an action in HA to switch to different screens (screen 1, screen 2, etc.). Another way could be (and this would be even more versatile) that I send a complete screen layout as JSON or YAML via an HA action or only updates (so that I can change only the control on position 3,2).

If you have decided which way to go, I could help develop this feature.

jtenniswood:
Sorry, this is a bit of an edge case, I haven't seen anyone else require this, so unlikely to include here.

PR #108, opened 2026-04-28
feat: allow partial button updates with keep/null semantics

ToBeHH:
Draft pull request to https://github.com/jtenniswood/espcontrol/issues/78

Still need to test it locally, but build is fine.


ToBeHH:
I tested it locally on my device, and things are working according to documentation. All checks are green. Converting it to a PR to be ready for review.

jtenniswood:
I don't think this is something that's in scope for this project. Thanks for the suggestion though.

PR #114, opened 2026-05-01
Add Hebrew glyph and right-to-left text support

yosef-chai:
## Summary

- Layer Heebo (the Roboto-companion family designed for Hebrew) on top of the existing Roboto label, clockbar and setup fonts via the font `extras` mechanism, so the Hebrew Unicode block is available in every UI surface that shows user-supplied text. Latin glyphs continue to come from Roboto, so existing Western entity labels are unchanged.
- Add `common/assets/hebrew_glyphs.yaml` covering the alef-tav letters and the five final forms, the niqud and Yiddish digraphs that Heebo ships, and the maqaf / geresh / gershayim punctuation. Cantillation marks Heebo does not include are intentionally omitted to keep the build clean.
- Enable LVGL's bidirectional text algorithm with `-DLV_USE_BIDI=1`. ESPHome leaves `LV_USE_BIDI` off by default, which makes Hebrew strings draw in storage order - so a Home Assistant entity named `שלום` rendered on screen as `םולש`. With BIDI on, LVGL's default `LV_BASE_DIR_AUTO` detects the script per label and flips the visual order automatically. Mixed Latin/Hebrew labels also lay out correctly.
- All four supported panels (4848S040, JC4880P443, JC1060P470, JC8012P4A1) get the Heebo extras on their device-specific `label_font_*` and `clockbar_font_*` overrides for parity.

Tested on a JC-S3-4848S040 with Hebrew-named entities; existing Latin-only setups are unaffected.

### DIFF
--- a/common/assets/fonts.yaml
+++ b/common/assets/fonts.yaml
@@ -42,12 +42,18 @@ font:
     size: 22
     bpp: 4
     glyphs: !include text_glyphs.yaml
+    extras:
+      - file: "gfonts://Heebo"
+        glyphs: !include hebrew_glyphs.yaml
 
   # Top bar: clock and temperature
   - file: "gfonts://Roboto"
     id: clockbar_font
     size: 20
     bpp: 4
+    extras:
+      - file: "gfonts://Heebo"
+        glyphs: !include hebrew_glyphs.yaml
 
   # Setup screens: titles (40px light)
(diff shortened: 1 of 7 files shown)


jtenniswood:
Very cool, thanks for that. Will take a look.

PR #123, opened 2026-05-05
Add light temperature slider

benJDtom:
## Summary

Adds a new card type — **Light Temperature Slider** — that controls a
Home Assistant light's `color_temp_kelvin` from a vertical slider.
Bottom of the slider is the coolest setting; top is the warmest.
Releasing the slider sends `light.turn_on` with `color_temp_kelvin`
(which also turns the light on if it was off).

## Settings

- **Entity ID** and **Min / Max Color Temp (K)** — defaults of 2000 K–6500 K cover most tunable white bulbs.
- **Use light color** — fills the bar with a colour lerped from warm amber to cool blue-white based on the current kelvin.
- **Show kelvin value when on** — replaces the card label with the live reading (e.g. `2500K`) whenever the light is on; reverts to the configured label / friendly name when off. Rounded to 50 K so the drag preview and HA's mired-quantized echo render identically.

## Behaviour

- Tapping the card without dragging is a no-op — only drag-and-release sends a command.
- When the light is **off**, the slider renders empty regardless of the last `color_temp_kelvin` reported by HA. Stale attribute updates while off are ignored.
- Echoes outside the configured display range are clamped so the slider position and the displayed kelvin stay in sync.
- Live drag updates the fill colour and the kelvin label without waiting for the HA echo.

## Test plan

- [ ] Card type appears as **Light Temperature Slider** in the dropdown
- [ ] Drag updates the bulb's colour temperature
- [ ] External HA changes update the slider position
- [ ] Light off → slider empty; light on → slider reflects current temp
- [ ] **Use light color** toggle works
- [ ] **Show kelvin value when on** toggle works (live during drag, reverts on off)
- [ ] Custom min/max range clamps correctly
- [ ] Works in subpages as well as the main grid


jtenniswood:
Thanks, I've updated and merged it, will test it before release.
Appreciate it!

PR #134, opened 2026-05-07
Notifications

wonderslug:
# Add HA-driven notification popup feature
IMG_1941

## Summary
This adds a top-layer LVGL overlay that Home Assistant can push notifications to via two ESPHome API actions.  The actions are ESPHome Native api actions to all devices for pushing notification popups from Home Assistant — one persistent and one auto-dismissing — with an optional title and a message ID for tracking. The popup renders as a full-screen overlay that wakes the screensaver and blocks interaction with the UI beneath. When dismissed, either by the user or by timeout, the device fires a corresponding event back to Home Assistant carrying the message ID so automations can tell the difference between a read and an unread notification.

### Two actions instead of one with an optional timeout field

ESPHome API action variables have no concept of defaults — every defined variable is required from HA's perspective. Using a single action with a timeout field meant users always had to supply it. Splitting into send_notification (no timeout) and send_expiring_notification (requires timeout) makes the intent explicit and removes ambiguity at the call site.

### Top-layer overlay instead of a new LVGL page

The notification renders on LVGL's top_layer, which sits above all pages. This means the popup appears correctly over the main grid, subpages, and the clock screensaver without any page navigation. The screensaver is woken automatically before the popup is shown.

### Touch blocking on the background

The full-screen scrim is clickable: true and consumes all touches, preventing accidental button presses on the main grid while reading a notification. Tapping anywhere — card or scrim — dismisses the popup.

### Distinct acknowledged vs. expired events

The device fires _esphome.notification_acknowledged_ on user tap and _esphome.notification_expired_ on auto-timeout. This lets HA automations distinguish between "user saw it" and "it timed out unread" — useful for critical alerts that should escalate if ignored.

## New API actions
* _send_notification_ — persistent, user must tap to dismiss

Variables: message, title , message_id

* _send_expiring_notification_ — auto-dismisses after timeout seconds

Variables: message, title, timeout (seconds), message_id

## HA events fired on close
_esphome.notification_acknowledged_ — user tapped to dismiss
_esphome.notification_expired_ — auto-timeout elapsed unread
Both events carry device_name and message_id for automation correlation.

### DIFF
--- a/common/device/screen_notification_popup.yaml
+++ b/common/device/screen_notification_popup.yaml
@@ -1,0 +1,133 @@
+# =============================================================================
+# SCREEN: NOTIFICATION POPUP
+# =============================================================================
+# Top-layer overlay for HA-pushed notifications. Two actions are exposed:
+#
+# Persistent (user must tap to dismiss):
+#   action: esphome.<device>_send_notification
+#   data:
+#     message: "Front door camera triggered"
+#     title: "Motion Detected"   # optional — omit or pass "" to hide title
+#     message_id: "motion_front"  # optional, returned in HA events
+#
+# Auto-dismissing:
+#   action: esphome.<device>_send_expiring_notification
+#   data:
+#     message: "Front door camera triggered"
+#     title: "Motion Detected"   # optional
+#     timeout: 15                # seconds until auto-dismiss
+#     message_id: "motion_front"  # optional, returned in HA events
+#
+# The popup fires one of two HA events on close:
+#   esphome.notification_acknowledged  – user tapped to dismiss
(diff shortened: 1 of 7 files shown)


wonderslug:
Not sure why the CI is failing on icons.  There is nothing in this that changes icons.

wonderslug:
closing to reopen and force recheck

wonderslug:
reopening to force recheck

heberlej (approved):

cybertimber2009:
Ooooh - notification support was something I was going to request be added.
Hoping to see this added!

wonderslug:
Closing this in favor of #1208.

PR #136, opened 2026-05-08
Timer card

### LINKED ISSUE #277: Timer card
Hi!

A native Timer Card to interface with Timer entities would be really nice! 

https://www.home-assistant.io/integrations/timer/

Some kind of setting the desired time, then showing the remaining time, and some sort of control (start, stop, etc.).

Thanks!

benJDtom:
## Summary
- Adds a Timer card that counts down a Home Assistant `timer.*` entity, gated behind the existing Developer / Experimental Features toggle.
- Tap-driven: idle/paused → `timer.start` (resumes if paused), active → `timer.cancel`, with optional two-tap confirmation prompt and configurable timeout (1–30s).
- Format auto-switches between `M:SS` and `H:MM` (drops seconds at ≥1h so long timers don't overflow the card; reverts below 1h).
- Highlights in primary color while active via `LV_STATE_CHECKED`.
- Holds `0:00` for ~5s after a natural finish then reverts to the configured duration; cancels revert immediately.
- 4Hz tick so the local countdown stays in phase with HA's `remaining` updates instead of skipping the last second.

## Implementation notes
- Added a small reusable `ConfirmationCtx` helper in `button_grid.h` so other destructive card actions can adopt the same two-tap flow.
- Subscribes to `state`, `duration`, and `remaining` attributes; anchors a local `remaining_anchor_ms` on each `remaining` push so the on-device interpolation doesn't drift.
- Distinguishes natural finish from cancel by checking estimated remaining at the `active → idle` transition.
- Uses single-letter compact code `Z` for `timer` to match existing card-type encoding.


ProfessorQuantumUniverse:
Hey, really cool! Exactly what I was looking for! Would be really cool to have it in the official Version!

jtenniswood:
Thanks for the submission, just updating it in a new PR #1935

PR #157, opened 2026-05-09
Add support for the Waveshare ESP32-S3 Touch LCD 7B

grizzlyjere:
This PR adds support for the Waveshare ESP32-S3 Touch LCD 7B which is 7" LCD with a resolution of 1024x600.

Developed with support of Codex, but tested on my personal device.

I didn't see a developer's guide so apologies in advance if I didn't follow the correct conventions.


jtenniswood:
Sorry I can't add device support if I can't test it, as it makes its impossible to validate new functionality without creating issues for users. I hope you understand.

PR #164, opened 2026-05-10
Add local action button type

wonderslug:
## Summary

- Adds a new `local` button type that calls a user-registered C++ callback by key name when tapped, bypassing Home Assistant entirely
- Register callbacks at boot with `register_local_action(key, label, fn)` — follows the same registry pattern as internal relays
- Works offline; no Home Assistant connection required
- Web UI card with free-text action key field and icon picker
- Documentation added under Card Types

## Use Cases

Any on-device action that doesn't need Home Assistant: IR blaster commands, GPIO pulses, UART commands to AV equipment, Wake-on-LAN, custom ESPHome component calls.

## Example

```yaml
esphome:
  on_boot:
    - priority: 700
      then:
        - lambda: |-
            register_local_action(
              "tv_off",
              "TV Off",
              [=]() { id(ir_blaster).transmit_nec(0x04FB, 0x08F7); }
            );
```

## Test plan

- [ ] Compile an ESPHome device config with a `register_local_action()` call in `on_boot`
- [ ] Configure a button as **Local Action** in the web UI with matching action key
- [ ] Tap the button — confirm the callback fires
- [ ] Configure a button with an unregistered key — confirm a warning is logged and no crash occurs


wonderslug:
Opened against wrong repo by mistake. Closing.

PR #165, opened 2026-05-10
Add local action button type

wonderslug:
## Local Action Button Type

Adds a `Local Action` button type that lets users call arbitrary on-device C++ callbacks directly
from the touchscreen, without routing through Home Assistant.

### How It Works

The implementation follows the same registry pattern already used by Internal Switches —
`register_internal_relay()` / `send_internal_relay_action()` in `button_grid.h`. This PR
extends that pattern beyond relay hardware to any user-defined callback:

```cpp
// Register in on_boot lambda
register_local_action("tv_off", "TV Off", [=]() {
    id(ir_blaster).transmit_nec(0x04FB, 0x08F7);
});
```
A local action card configured with the matching key dispatches to the registered callback
on tap. If no action is registered for the key, a warning is logged and nothing else happens.

The entity config field carries the action key, consistent with how internal cards use
it to store the relay name. No new config fields are needed.

### Why
Some useful actions belong entirely on the device: IR blaster commands, GPIO pulses, UART
commands to AV equipment, Wake-on-LAN packets, ESPHome script calls. These don't need Home
Assistant in the loop, and should work even when HA is offline. The local type gives users
a clean way to wire these up from the web UI without modifying the core firmware.

#### Bigger Picture
ESPHome is a capable platform — panels running EspControl have access to I2C, UART, IR,
GPIO, and any other component the ESP32 supports. For the most part, EspControl has been
primarily a polished frontend for Home Assistant, with the panel acting as a thin UI layer
that passes every interaction upstream.

Local Actions can give EspControl more depth. With this pattern, the panel can directly control
hardware connected to it — making it a proper controller in its own right, not just a
display. Home Assistant remains the hub for orchestration and state, but the panel gains
agency: it can act immediately, locally, and independently when that's the right model for
the task. 

This is the same philosophy behind the existing Internal Switches type — the relay variant
proved the pattern works. Local actions generalizes it to the full ESPHome component ecosystem.

Changes
_components/espcontrol/button_grid.h_ — LocalActionControl struct, registry, register_local_action(), send_local_action(), dispatch branch and push-style visual treatment
_src/webserver/types/local.js_ — web UI card type with action key field and icon picker
_docs/card-types/local-actions.md_ — user-facing documentation


wonderslug:
closing in favor of rebased and cleaned up #178

PR #177, opened 2026-05-11
Add camera popup

snazy2000:
Adds a fullscreen camera snapshot overlay that can be triggered from any button card by setting the entity to a camera entity. Snapshots are fetched from HA's camera proxy REST endpoint using a Bearer token which is set in the settings

Features:
- Periodic JPEG snapshot refresh (configurable interval)
- Loading spinner on first open; silent refresh on subsequent frames
- Bearer token set via existing HA Token web UI field

I have only tested this on the 4848s040 but it works so well!

AI Disclosure, Claude helped put this together but all code has been looked over and the feature has been tested best I can with devices I have.


jtenniswood:
Thanks for sharing this, I'm adding image cards now, using you code as inspiration rather than directly. Appreciate the work!

jtenniswood:
I think this has now be replaced by the camera card recently been added.

PR #178, opened 2026-05-11
Add local action support with registration and handling

wonderslug:
# Local Action Button Type

Adds a `Local Action` button type that lets users trigger any on-device behaviour directly from the touchscreen — no Home Assistant required. The action can be anything that runs in an ESPHome lambda: calling a YAML-defined script, transmitting IR codes, pulsing a GPIO, sending UART commands, or raw C++ against any ESPHome component.

## How It Works

The implementation follows the same registry pattern already used by Internal Switches — `register_internal_relay()` / `send_internal_relay_action()`. This PR extends that pattern beyond relay hardware to any user-defined action:

```yaml
esphome:
  on_boot:
    - priority: 700
      then:
        - lambda: |-
            // Call a YAML script — no C++ required
            register_local_action("night_mode", "Night Mode", [=]() {
                id(night_mode_script).execute();
            });

            // Or go straight to hardware
            register_local_action("tv_off", "TV Off", [=]() {
                id(ir_blaster).transmit_nec(0x04FB, 0x08F7);
            });
```

The lambda body is an ESPHome lambda — the same syntax used in any `on_boot`, `on_press`, or automation `lambda:` block. Anything valid there is valid here: `id()` references, script calls, component methods, Home Assistant service calls, or plain C++. The callback runs on the device the moment the card is tapped.

Screenshot 2026-05-11 at 8 49 02 PM

A local action card configured with the matching key dispatches to the registered callback on tap. If no action is registered for the key, a warning is logged and nothing else happens.

The entity config field carries the action key, consistent with how internal cards use it to store the relay name. No new config fields are needed.

## Action Key Auto-Population

When a user opens the settings panel for a Local Action card, the web UI fetches `GET /local_actions` from the device — an endpoint served directly by the firmware that serializes the live action registry as JSON:

```json
[{"key": "tv_off", "label": "TV Off"}, {"key": "unlock_door", "label": "Unlock Door"}]
```

The panel shows a dropdown of all registered actions. Selecting one auto-fills the card's Label field from the registered label. Three states are handled:

- **Actions available** — dropdown picker with key + label pairs
- **No actions registered** — banner prompting the user to add `register_local_action()` calls
- **Device unreachable** — error banner with a manual key entry field as fallback (since the action may be registered but the device can't be reached for discovery)

The endpoint is registered automatically via `core_infra.yaml`'s `on_client_connected` hook — no per-device wiring required. It uses `global_async_web_server()` added to `web_server_idf`.

## Why

Some useful actions belong entirely on the device: IR blaster commands, GPIO pulses, UART commands to AV equipment, Wake-on-LAN packets, ESPHome script calls. These don't need Home Assistant in the loop, and should work even when HA is offline. The local type gives users a clean way to wire these up from the web UI without modifying the core firmware.

### Bigger Picture

ESPHome is a capable platform — panels running EspControl have access to I2C, UART, IR, GPIO, and any other component the ESP32 supports. For the most part, EspControl has been primarily a polished frontend for Home Assistant, with the panel acting as a thin UI layer that passes every interaction upstream.

Local Actions can give EspControl more depth. With this pattern, the panel can directly control hardware connected to it — making it a proper controller in its own right, not just a display. Home Assistant remains the hub for orchestration and state, but the panel gains agency: it can act immediately, locally, and independently when that's the right model for the task.

This is the same philosophy behind the existing Internal Switches type — the relay variant proved the pattern works. Local Actions generalises it to the full ESPHome component ecosystem.

## Changes

- `components/espcontrol/button_grid_config.h` — `LocalActionControl` struct, registry, `register_local_action()`, `send_local_action()`, `LocalActionHandler` (serves `GET /local_actions`), `register_local_action_endpoint()`
- `components/espcontrol/button_grid_cards.h` — `setup_local_action_card()` with push-style visual treatment
- `components/espcontrol/button_grid_actions.h` — dispatch branch in `handle_button_click()`
- `components/espcontrol/button_grid_grid.h` — visual dispatch in `setup_card_visual()`, subscription skip in `grid_phase2()`, subpage click handler
- `components/web_server_idf/web_server_idf.h/.cpp` — `global_async_web_server()` accessor (shared infrastructure, also used by local-sensors)
- `common/device/core_infra.yaml` — auto-wires `register_local_action_endpoint()` on client connect
- `src/webserver/types/local_actions.js` — web UI card type: action dropdown with auto-fill, empty/unreachable fallback states, icon picker
- `docs/card-types/local-actions.md` — user-facing documentation


wonderslug:
Let me give a practical example of using this.  Using the local actions I was able to setup my ESPControl device to act as a BLE keyboard and send key combinations I need directly to the machine its connected to.  I use this to control Zoom mute and video and any other key combos.

IMG_1952


In order to do this added an ESP32-C3 SuperMini as a coprocessor to my guition-esp32-s3-4848s040.  I was able to communicate via the UART between the two controllers, just like the guition-esp32-p4-jc1060p470 does to the C6 it uses for wifi.  

IMG_1951

This allows me to program the ESPHome to setup all the commands to communicate with saying what key combos to send.  This lets ESPControl become more of a controller for the ESP device its connected to and do more with it right on the board without having to go through Home Assistant.  It allows it to do things that Home Assistant just cant do.

jtenniswood:
Thanks for that, could you check that the merged version works as expect. Cheers!

wonderslug:
Validated against main, works fine. thanks for getting it in.

Just for future, quickest way to test is just setup a log message for the action.

```
  on_boot:
    - priority: 700
      then:
        - lambda: |-
            register_local_action(
              "debug_test",
              "Debug Test",
              [=]() { ESP_LOGD("local_action", "debug_test fired"); }
```

jtenniswood:
Awesome, many thanks

PR #179, opened 2026-05-12
Add local sensor support and related UI components

wonderslug:
# Add Local Sensor card type

Adds a new **Local Sensor** card that displays live readings from any ESPHome `sensor` or `text_sensor` component running directly on the device, with no Home Assistant dependency.

---

### Numeric mode
Shows a large value + unit + label — same visual layout as the existing HA-backed Sensor card, but sourced locally. The display updates automatically whenever the sensor fires its `on_value` callback; no polling, no HA subscription.

Screenshot 2026-05-11 at 8 02 28 PM

### Text mode
Shows a chosen icon and renders the live `string` state where the button label normally appears — useful for template text sensors like door state, AQI category, occupancy mode, etc.

Screenshot 2026-05-11 at 8 03 32 PM

Both modes are read-only; tapping does nothing (clicks are suppressed in `handle_button_click`).

---

### `components/web_server_idf/web_server_idf.h` + `.cpp`
Adds `global_async_web_server()` — an inline singleton accessor that stores the `AsyncWebServer*` when the server starts. This lets other components register HTTP handlers without holding a direct reference to the web server component.

### `components/espcontrol/button_grid_config.h`
- Adds `#ifdef`-guarded includes for `sensor/sensor.h`, `text_sensor/text_sensor.h`, and `web_server_idf.h` at the top of the file.
- Defines `LocalSensorControl` struct (key, is_text flag, precision, LVGL label pointers) and `local_sensor_registry()` singleton.
- `LocalSensorHandler` — an `AsyncWebHandler` subclass that serves `GET /local_sensors` as a JSON array. Each entry has `key`, `name`, `unit`, `type` (`"numeric"` or `"text"`), and an optional `"internal": true` flag. Iterates `esphome::App.get_sensors()` and `get_text_sensors()` inside `#ifdef USE_SENSOR` / `#ifdef USE_TEXT_SENSOR` guards so the handler compiles cleanly on configs without those components.
- `register_local_sensor_endpoint()` — idempotent (static-bool guarded); fetches the server pointer from `global_async_web_server()` and registers `LocalSensorHandler`. Called from `on_client_connected` in `core_infra.yaml`.

### `components/espcontrol/button_grid_cards.h`
- `send_local_sensor_update(key, float)` / `send_local_sensor_update(key, const char*)` — public API for pushing a value to a registered card from an ESPHome lambda. Walks `local_sensor_registry()` and sets the LVGL label text directly. Lives here (not in `button_grid_config.h`) because it calls `set_wrapped_button_label_text`, which is defined in `button_grid_layout.h` (included between config and cards).
- `setup_local_sensor_card()` — LVGL visual setup. Marks the button non-clickable, populates the sensor container or text label, upserts into `local_sensor_registry()`, then auto-subscribes to the matching ESPHome sensor's `add_on_state_callback` so the display updates live. Falls back gracefully if the named sensor isn't found (the `send_local_sensor_update` push path still works).

### `components/espcontrol/button_grid_actions.h`
`local_sensor` added to the early-return guard in `handle_button_click` alongside `sensor`, `text_sensor`, `calendar`, etc. — prevents spurious click handling.

### `components/espcontrol/button_grid_grid.h`
- `setup_card_visual`: dispatches `local_sensor` type to `setup_local_sensor_card`.
- `grid_phase2` (main grid loop): `local_sensor` gets a `continue` — no HA subscriptions needed.
- `grid_phase2` (subpage loop): same `continue` for subpage slots.

### `common/device/core_infra.yaml`
`register_local_sensor_endpoint()` appended to the existing `on_client_connected` lambda. This fires every time an HA client connects, but the function is idempotent so the handler is only ever registered once. No `on_boot` or web server `id` needed — the `global_async_web_server()` pointer covers it.

---

### `src/webserver/types/local_sensor.js`
New card type registered as `"local_sensor"`. Settings panel:
- **Display** segment toggle (Numeric / Text) — switches the visible field sections and updates `precision` in the config string.
- **Sensor dropdown** — populated by `GET /local_sensors`; filtered to `"numeric"` or `"text"` entries matching the current mode. Selecting an entry auto-fills Label and Unit from the device metadata if those fields are empty.
- **Show internal sensors** toggle — reveals diagnostic entities (WiFi RSSI, uptime, etc.) that are filtered out by default.
- **Manual fallback** — if the fetch fails (device unreachable from the browser), an error banner is shown and a free-text **Sensor Key** field appears so the card can still be configured offline.
- `renderPreview`: Numeric mode shows `0` / `0.0` / `0.00` sample with unit badge; Text mode shows the chosen icon with a `"State"` placeholder label.

### `src/webserver/modules/controls.js`
`local_sensor` added to the sensor-background-color check in the grid preview builder (line 1301), so the card renders with the tertiary/sensor colour in the web UI preview — consistent with `sensor`, `weather`, `calendar`, etc.

---

## Documentation

`docs/card-types/local-sensors.md` — new page covering:
- Numeric and Text modes with setup steps
- Internal sensors toggle
- Device-unreachable fallback
- How auto-subscribe works and what `--` means at boot
- A sensor/use-case reference table
- `send_local_sensor_update()` escape hatch with YAML examples
- When to use Local Sensor vs the HA-backed Sensor card

`docs/.vitepress/config.mts` — nav entry added after "Internal".


jtenniswood:
Thanks, I've merged it in

PR #244, opened 2026-05-24
Added tumble-dryer-off icon

pvdl:


jtenniswood:
Thanks, I've added the icon directly, as there's script to integrate it in that runs after commit. It's easy for me to add them in, just raise an issue.

PR #286, opened 2026-05-31
Use generic firmware font styles

jtenniswood:
## Summary
- Replace physical firmware font IDs with generic style names such as font_text_body, font_text_title, font_number_value, and font_icon_main.
- Remove the shared common font package from normal device builds so each device defines the same generic names with device-specific sizes.
- Consolidate similar setup, label, climate, media, icon, and number font roles, including numeric-only modal number glyph coverage where practical.

## Font impact
- Runtime font definitions are reduced on normal display devices by removing duplicate shared setup fonts.
- Approximate loaded font count changes: JC106 14 -> 10, JC488 18 -> 8, JC801 17 -> 10, ESP32-P4-86 18 -> 10, S3 480x480 15 -> 10, TRMNL 4 -> 4.

## Validation
- python3 scripts/generate_device_slots.py --check
- npm run check:product
- npm run check:firmware-modals
- python3 scripts/build.py icons --check
- Compiled representative factory builds for all current devices: JC106, JC488, JC801, ESP32-P4-86, S3 480x480, and TRMNL.


JimBuzbee:
Saw this PR and wondered if it would affect a need/desire I have - My use-case is to use one of the 4x4 displays as just an outside temperature display. Would be nice to use a very large font with a large tile for just temperature and maybe time i.e. the temperature display would mostly fill up the entire screen of a 4x4 device.

PR #302, opened 2026-06-04
Add Dutch language

hunter-nl:
I've added Dutch language translation.


common/config/strings.nl.txt
```diff
@@ -0,0 +1,140 @@
...
+hail=Hagel
+hazy=Nevelig
+heat=Verwarmen
+heat_cool=Verwarmen/Koelen
+heating=Verwarmen actief
+heavy_snow=Zware sneeuwval
+high=Hoog
+high_target=Hoog doel
+home=Thuis
+hourly=Per uur
```
michamichamicha: Since this is used on the Update frequency feature, maybe it makes sense to use `Elk uur` instead?
Also noticed the `daily` (`Dagelijks` in Dutch) key is missing here, but also in the English source file? @jtenniswood

jtenniswood: Blimey, that was quick! I only added it last night!
Thanks for that, will bring that in now

hunter-nl: @michamichamicha You're correct. I will update hourly to "Elk uur". I can not find if "daily" is in use. Maybe @jtenniswood can confirm this.

hunter-nl:
@jtenniswood I see that you just added NL and other languages. Then this PR is not needed anymore :-)

common/config/strings.nl.txt
jtenniswood: I don't believe so, I've limited this to just words that are hardcoded into the device display.
I've left the web server words out, as it was considerably more, and I thought this would be admin facing, rather than all user facing. 
Any words that aren't covered should be coming from HA (in theory)

jtenniswood:
Sorry, I didn't see the PR before I generated the others. No idea if the content is correct though.

hunter-nl:
It looks good to me. Only ```still=Stil```. If that can be changed into ```still=Vast```. It is used in context of a Fan oscillating or still. Current "Stil" means in Dutch silent and that is not what it means to be.

jtenniswood:
Great, I've applied that change

hunter-nl:
Thanks! I'm just flashed it to test your new changes. So far so good.

PR #347, opened 2026-06-07
feat: add sensor_chart card type (live value + LVGL sparkline history)

fbeauchamp:
## Summary

PXL_20260607_123852817 PORTRAIT

Adds a new `sensor_chart` card type that displays a live numeric sensor value
alongside a scrolling sparkline history chart built on LVGL's `lv_chart`
widget. The chart accumulates HA state updates over a configurable interval
and plots one aggregated point (avg / max / min / last) per interval.

### C++ side

**`button_grid_config.h`** — new `sensor_chart` config helpers:
- `sensor_chart_normalize_options()` — round-trips `interval`, `mode`,
  `points` from the stored options string
- `sensor_chart_interval_ms()`, `sensor_chart_mode_index()`,
  `sensor_chart_point_count()` — accessors with safe defaults

Two new fields added to `BtnSlot`: `chart` (`lv_obj_t*`) and `chart_series`
(`lv_chart_series_t*`), both initialised to `nullptr`.

**`button_grid_cards.h`** — `setup_sensor_chart_card()` creates the
`lv_chart` widget at phase-1 grid setup. Key decisions:
- `LV_CHART_TYPE_LINE` + `LV_CHART_UPDATE_MODE_SHIFT`: new points push the
  history left
- All points initialised to `LV_CHART_POINT_NONE` so the line only appears
  where real data exists
- `lv_obj_move_foreground(chart)` is required: the sensor text label uses
  `LV_OBJ_FLAG_OVERFLOW_VISIBLE` + wrapping, which causes LVGL to place it
  above later-created siblings in the draw order. Without the foreground
  call the chart is hidden behind the label.
- Line colour is always `DEFAULT_CHART_COLOR` (0xFF8C00). Do **not** use
  `palette.sensor_val` — that is the button background colour and makes the
  line invisible.
- `lv_obj_set_style_size(chart, 0, 0, LV_PART_INDICATOR)` — LVGL 9 requires
  **four** arguments here (width, height, selector); three-arg calls silently
  do nothing and leave visible data-point dots.

**`button_grid_subscriptions.h`** — `SensorChartAcc` accumulator + callback:
- On the **first** HA state update: all existing chart points are pre-filled
  with the current value (`lv_chart_set_all_value`) and the Y range is
  centred on it. This avoids an initial flat line at 0.
- The "already seeded" state is tracked with a `bool seeded` flag on the
  accumulator struct. Do **not** use `acc_count == 1` for this check:
  `acc_count` resets to 0 after every flush, so that condition would
  re-seed (and erase history) on every single post-flush update.
- On each flush: one aggregated point is pushed, and the Y axis range is
  auto-scaled to `[data_min − margin, data_max + margin]`.
- The accumulator is heap-allocated (`new SensorChartAcc{}`) and intentionally
  not freed — consistent with the existing pattern for other subscription
  contexts in this codebase.

**`__init__.py`** — calls `add_lv_use("CHART")` at import time to set
`LV_USE_CHART 1` in the generated `lv_conf.h`. Without this, ESPHome's LVGL
component leaves the chart widget disabled and `lv_chart_create()` returns
`nullptr` silently at runtime.

**`button_grid_grid.h`** — `find_chart_in_btn()` helper: phase-2 recovers
the chart widget by walking the button's LVGL children, because phase-1 and
phase-2 each build a fresh local `BtnSlot` array. The `s.chart` pointer set
in phase-1 is gone by phase-2, so child-scanning is the only reliable way to
find the widget.

### JS / web UI side

**`src/webserver/types/sensor_chart.js`** — new card type registration with:
- Interval select (1 s test / 5 min / 15 min / 30 min / 1 h / 2 h / 6 h /
  12 h / 24 h)
- Mode segment (Avg / Max / Min / Last)
- Points field (2–60, default 24)
- SVG sparkline preview in the web UI editor

**`src/webserver/modules/config_codec.js`**:
- `normalizeButtonConfig`: `sensor_chart` branch preserves options unchanged
  (the surrounding else-if chain would otherwise clear them)
- `buttonConfigFields`: `sensor_chart` excluded from the options-clearing
  condition, consistent with `action`, `garage`, etc.

> **Note**: the subpage serialization path (`serializeLegacySubpageConfig` /
> `serializeCompactSubpageConfig`) also required the same `sensor_chart`
> exclusion. In the current codebase both functions delegate to
> `buttonConfigFields` via `subpageLegacyButtonFields` /
> `subpageCompactButtonFields`, so the fix is in one place. An earlier
> version of this code had the exclusion missing from the inline subpage
> serializers, causing options to silently reset to defaults for cards placed
> inside subpages.

## Known limitations / not included

- **Area fill** (`LV_PART_ITEMS` background): not rendered on this
  display/LVGL build regardless of `bg_opa` value. The fill code was removed;
  the root cause was not investigated.
- The `"1 s (test)"` interval option is intentionally left in the UI for
  development convenience. It should be removed before a stable release.
- Y-axis auto-scaling uses a fixed ±2-unit margin around the data range —
  adequate for typical sensor values but not ideal for very noisy or
  near-zero signals.

## Testing

Tested on JC3248W535 (ESP32-S3, LVGL 9) with a power sensor updating every
few seconds. Chart history, Y-axis scaling, interval/mode/points persistence
(main screen and subpages), and device reboot all verified.

### DIFF
--- a/components/espcontrol/__init__.py
+++ b/components/espcontrol/__init__.py
@@ -10,6 +10,16 @@ import esphome.codegen as cg
 import esphome.config_validation as cv
 import os
 
+# Enable LV_USE_CHART in the generated lv_conf.h so that sensor_chart cards
+# can create lv_chart sparkline widgets at runtime.  This must run at import
+# time (before any to_code() is called) so that the LVGL component sees the
+# flag when it writes lv_conf.h.
+try:
+    from esphome.components.lvgl.helpers import add_lv_use
+    add_lv_use("CHART")
+except Exception:
+    pass
+
 CODEOWNERS = ["@jtenniswood"]
 
 CONF_ACTION_RESPONSES = "action_responses"
--- a/components/espcontrol/button_grid_cards.h
+++ b/components/espcontrol/button_grid_cards.h
@@ -678,3 +678,62 @@ inline void setup_subpage_parent_state_card(BtnSlot &s, const ParsedCfg &p,
     s, subpage_chevron_enabled, subpage_chevron_x, subpage_chevron_y,
     subpage_chevron_text_width_percent);
 }
+
+// ── Sensor chart card ──────────────────────────────────────────────────────────
+// Like sensor card, but adds an lv_chart sparkline at the bottom of the button.
+// Chart height and bottom margin are fixed in pixels; the value + label use the
+// same positions as a regular sensor card.
+
+constexpr int SENSOR_CHART_HEIGHT_PX = 40;
+constexpr int SENSOR_CHART_BOTTOM_MARGIN_PX = 16; // gap for text_lbl
+
+inline void setup_sensor_chart_card(BtnSlot &s, const ParsedCfg &p,
+                                    bool has_sensor_color, uint32_t sensor_val) {
+  if (has_sensor_color) {
+    lv_obj_set_style_bg_color(s.btn, lv_color_hex(sensor_val),
+      static_cast<lv_style_selector_t>(LV_PART_MAIN) | static_cast<lv_style_selector_t>(LV_STATE_DEFAULT));
+  }
+  lv_obj_clear_flag(s.btn, LV_OBJ_FLAG_CLICKABLE);
+  lv_obj_add_flag(s.icon_lbl, LV_OBJ_FLAG_HIDDEN);
+  lv_obj_clear_flag(s.sensor_container, LV_OBJ_FLAG_HIDDEN);
+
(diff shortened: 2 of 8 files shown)


fbeauchamp:
fixed, the issue was a memory induced error on my small device

jtenniswood:
That's very cool, like it!

jtenniswood:
This is interesting but I don't think its the right integration at this point.
Appreciate the input

PR #351, opened 2026-06-07
SenseCAP Indicator D1

davidmerrique:
## Summary
- add Seeed SenseCAP Indicator D1 device/build definitions
- add generated docs, device profile metadata, and webserver assets for the new display
- update compatibility/device slot generation to include the D1 profile

Product: https://www.seeedstudio.com/SenseCAP-Indicator-D1-p-5643.html

## Tests
- npm run check:fast
- npm run check:web-browser-smoke


jtenniswood:

davidmerrique:
@jtenniswood That's fine. Do you have a plan for community devices? 

It'd be great if we could just supply a package like this:
```yaml
packages:
  setup:
    url: https://github.com/davidmerrique/espcontrol-sensecap-indicator-d1/
    file: espcontrol-sensecap-indicator-d1/packages.yaml
    refresh: 1sec
```

But it seems like `www.js` has a baked-in device config. If we could decentralize device profiles out of mainfest.json, then perhaps a community package can contain it's own device profile.

lamiskin:
Hi @davidmerrique — heads up that I've adapted your SenseCAP Indicator D1 port into the [community devices repo](https://github.com/lamiskin/espcontrol-community-devices), credited to you, building against current EspControl with a [browser installer](https://lamiskin.github.io/espcontrol-community-devices/screens/seeed-sensecap-indicator-d1). Thanks for the original work! It's listed as *Untested* since I don't have the hardware — if you ever flash it and it works, a note on the community tracker would let me mark it verified.

lamiskin:
Following up on my note above — the SenseCAP Indicator D1 is now **hardware-verified**. @mglev1n flashed the community build on a real panel and confirmed it running: home screen renders, cards are live and pulling entity state (climate, weather, clock). Photo in the verification report: lamiskin/espcontrol-community-devices#62

So the device has moved from *Untested* to **Working** on the community tracker:
https://lamiskin.github.io/espcontrol-community-devices/screens/seeed-sensecap-indicator-d1

Thanks again @davidmerrique for the original port — it's credited to you on that page, and anyone with this panel can now flash it from the browser installer without the "nobody has tried this" caveat.

To be clear for anyone finding this thread: that's the *community* repo, which is unofficial and unsupported by @jtenniswood — issues with it belong on the community tracker, not here.

PR #362, opened 2026-06-08
Fix UTF-8 corruption when saving subpage sensor config with non-ASCII units

AndreasUfert:
Two bugs caused silent data corruption for users who configure sensor cards with units like °C, μg/m³ or similar non-ASCII strings via the WebUI:

1. decodeConfigField used String.fromCharCode(parseInt(hex, 16)) to decode each %XX escape individually. This treats every byte as a Unicode code point, so a UTF-8 multi-byte sequence like %C2%B0 (°) was decoded as two separate Latin-1 characters Â and ° instead of the single character °. Fix: use decodeURIComponent() which understands UTF-8 byte sequences and decodes consecutive %XX runs into the correct Unicode characters.

2. splitSubpageConfigChunks split the config string on CHARACTER position 255, but the ESPHome text entity enforces max_length=255 in BYTES. A chunk of 255 characters containing multi-byte UTF-8 characters (e.g. ° = 2 bytes, μ = 2 bytes) can exceed 255 bytes and is then silently rejected by the device (returns HTTP 200 but discards the write), causing the subpage config to revert to the previous value without any error shown to the user. Fix: add utf8ByteLength() and rewrite the splitter to track byte count and split on character boundaries so each chunk stays within the byte limit.

### DIFF
--- a/src/webserver/model/card.ts
+++ b/src/webserver/model/card.ts
@@ -85,8 +85,16 @@ export function encodeConfigField(value: string | number | boolean | null | unde
 }
 
 export function decodeConfigField(value: string | number | boolean | null | undefined): string {
-  return String(value || "").replace(/%([0-9a-fA-F]{2})/g, (_match, hex: string) => {
-    return String.fromCharCode(parseInt(hex, 16));
+  const str = String(value || "");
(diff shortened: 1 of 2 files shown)


jtenniswood:
Thanks, appreciate it

PR #365, opened 2026-06-09
feat: support labels on text-mode sensor cards

### LINKED ISSUE #296: Labels for Text Sensors
Numeric Sensors can have a label, text sensor not. That makes it difficoult to identify if you have several text sensors on the same page. The possibility to add a label would be a plus.

mvanhorn:
## Summary

Adds the **Label** field to text-mode sensor cards in the web setup UI, so multiple text sensors on the same page can be told apart.

## Why this matters

Per #296, numeric sensor cards already support a custom label but text-mode sensor cards do not, making several text sensors on one page hard to distinguish. The firmware already renders a non-empty `p.label` for sensor cards (`components/espcontrol/button_grid_cards.h`) regardless of mode, so this was a web-only gap: the setup UI never exposed the field in text mode, and actively cleared any existing label on switch.

## Changes

- `src/webserver/types/sensor.js`:
  - Moved the **Label** field out of the numeric-only section into its own section that is visible in both **Numeric** and **Text** modes (still hidden in **Icon** mode, matching existing behavior).
  - Removed the text-mode reset that wiped `b.label` (and persisted an empty label) when switching numeric to text, so an existing label now survives the switch.
  - Existing `rerender`/`saveField` wiring is preserved, so the label persists through the webserver config round-trip.
- Regenerated the bundled `docs/public/webserver/*/www.js` outputs so the served assets carry the fix.

## Testing

- `python3 scripts/build.py --check` - all generated outputs up to date.
- `node --check src/webserver/types/sensor.js` - passes.
- `npm run check:types` (`tsc --noEmit`) - passes.
- Manual reasoning against the firmware render path: text-mode cards display the label in `s.text_lbl` and the state in `s.sensor_container` (separate LVGL objects), so the persisted label now renders on-device.

Note: `npm run check:web-smoke` currently fails on an unrelated, pre-existing assertion (`date preview follows selected language month names`) that also fails on a clean `main` checkout under Node v26; it is not touched by this change.

Fixes #296

AI was used for assistance.


jtenniswood:
I would rather keep it simple and allow only labels or status

timothyBoye:
@jtenniswood this would be super useful for me as I have some text based sensor helpers that report humidity/air quality etc as "Good" or "Excellent" etc, and so currently it is anyone's guess as to which room and sensor type they are reporting. The icons help for the later or the former but it hard to come up with an icon to represent both so a label option would be great. If you keep the label field optional people could keep the text sensors as they are now or choose to add a label?

PR #475, opened 2026-06-13
Fix media standby when cover art is disabled

### LINKED ISSUE #472: BUG: Screen Saver stop working
I don't know what happened...
 I installed version 2.2.9 and everything worked on my 4" display.
So I tried the CoverArt option, but it didn't show the disc covers and caused other display problems (I'll investigate the type of problems later). 

So I decided to turn the cover art option off, now the screen saver no longer activates (Timer screen saver, after 10 seconds showing the clock).

jtenniswood:
## Purpose
Fixes #472.

When Media Cover Art is turned off, the display should return to the normal screensaver/standby behavior. Previously, the separate media playback sleep-prevention switch could remain active, so a playing media player kept the main screen awake even though cover art was no longer shown.

## Practical impact
- Turning off Media Cover Art now also turns off Media Player Sleep Prevention on the device.
- The web settings page immediately unticks the wake-lock option when Media Cover Art is disabled.
- Generated web UI bundles were refreshed for all supported display profiles.
- A firmware regression check now verifies that disabling cover art clears media sleep prevention.

## Checked
- npm run check:firmware-ha-bindings
- npm run check:web-smoke
- npm run check:types
- npm run check:generated

## Device testing
Flash this branch to a display with Media Cover Art configured. Start playback on the selected media player, turn Media Cover Art off, leave the player in the playing state, and confirm the display still goes to standby after the configured screensaver timeout.


Skyman81:
as mentioned in another post:

with 2.2.10
it's working 50% I mean:

When the cover art option on displays is DISABLED now all displays enter screen saver mode without problems. DONE.

When the cover art option display in ENABLED:
On the ESP32P4 display everything is working good!
THe Esp32S3 4" the display stay active in home with no screensaver.
You need to touch the screen once for the screen saver to work again after the time

Issue #473, opened 2026-06-13
Artwork pulling assumes that Home Assistant is running on port 8123 and using http


salnajjar:
My artwork doesn't display using my Home Assistant install which is using https and served over the standard https TCP port of 443

In these two files in the code base:
```
common/device/core_infra.yaml (line 33)
common/device/screen_cover_art.yaml (line 124)
```

The URL constructor is hard coded to use http and port 8123:
`"http://" + host + ":8123"`

Looking in the device logs, when it tries to pull the artwork, it's going to:
`http://192.168.1.2:8123/api/media_player_proxy/media_player.espcontrol_p4_86_voice_media_player`
which obviously doesn't work on any customised HA install.

jtenniswood:
Thanks, will take a look at adding advanced setup options

jtenniswood:
Thanks for reporting this. I’ve opened a draft fix in #478 that makes the artwork protocol and port configurable instead of hardcoding `http://<host>:8123`.

For your setup, the branch should let you set:
- Home Assistant Artwork Protocol: `https`
- Home Assistant Artwork Port: `443`

Once you have a chance to test the PR build on a display, please leave feedback here and let us know whether it resolves the artwork loading issue for you.

jtenniswood:
I've made an update, and doing a release now (2.2.10). If you could test if and let me know the results. Thanks

jtenniswood:
Let me know if that solved the issue

salnajjar:
@jtenniswood I had to add an entry in the web ui to access the new settings, but once I had done that, and updated the config to use https its all been working perfectly.

jtenniswood:
Great, thanks for the feedback

PR #478, opened 2026-06-13
Fix Home Assistant artwork URL settings

jtenniswood:
## Summary
- Adds device web settings for the Home Assistant artwork protocol and port.
- Keeps the existing default as `http` on port `8123`, so current installs should behave the same after update.
- Uses the selected protocol and port when building the artwork proxy base URL, instead of hardcoding `http://<host>:8123`.

## Why
Some Home Assistant installs serve HTTPS on port 443 or use another custom port. The artwork downloader was always trying `http://<host>:8123`, which meant media artwork could not load on those setups.

## Testing
- `npm run check:fast`
- Docker ESPHome compile: `builds/esp32-p4-86.factory.yaml`

## Device testing notes
Flash this branch to an affected display, then set:
- Home Assistant Artwork Protocol: `https`
- Home Assistant Artwork Port: `443`

After applying/rebooting, start media playback and confirm the cover artwork loads.

Related to #473

### DIFF
--- a/common/device/core_infra.yaml
+++ b/common/device/core_infra.yaml
@@ -30,7 +30,9 @@ api:
             } else if (host.find(':') != std::string::npos && host.front() != '[') {
               host = "[" + host + "]";
             }
-            id(cover_art_home_assistant_base_url) = "http://" + host + ":8123";
+            int port = static_cast<int>(id(cover_art_home_assistant_artwork_port).state);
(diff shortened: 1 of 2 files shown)


salnajjar:
@jtenniswood I had to manually expose the connectivity options in the UI, but the settings themselves all work and cover art now displays perfectly. (Let me know if you want a diff for the 4 files I updated to have it displayed)

PR #551, opened 2026-06-17
Added ESP32-C6 firmware update to ESP32-P4-86 device

hunter-nl:
## Summary
Due stability issues with the fabric firmware in the ESP32-C6 on Waveshare ESP32-P4-86 device, I decided to add updating firmware of this co-processor.

Added buttons to check and update firmware of the co-processor ESP32-C6 on the ESP32-P4 device.
It expose (every minute) to HA the current, latest firmware versions and if update is available.
The updates manifest will be downloaded every 24h automatically. The check button does this manually.

Diagnostic

Configuration

## Testing

- [X] I checked the change locally where practical.
- [X] Affected display/device, if applicable:
ESP32-P4-86

## Notes for testing
The fabric Waveshare ESP32-P4-86 display has an old ESP32-C6 firmware. It has WiFi stability issues, like lost connection randomly. It also does not expose current firmware version.
So the first time running with included this PR, the current firmware version will be `Unknown` shown in the entity. After the display is started and wifi is connected, wait one minute before you press on the ESP32-C6 firmware update button. It does not harm is you do. It must fetch first the manifest to know the [available firmwares](https://esphome.github.io/esp-hosted-firmware/).
After pressed on the ESP32-C6 firmware update button, just wait. The device will restart automatically when ESP32-C6 firmware has been updated.

The ESPhome update process will determine the correct latest **compatible** available firmware. There are newer versions available, but the ESPhome and esp-idf framework used must be compatible with the hosted ESP32-C6 firmware.

Debug logging shows information about the ESP32-C6 update info/processing.
-

## Issue handling
No issue created.

### DIFF
--- a/devices/esp32-p4-86/device/network_coprocessor.yaml
+++ b/devices/esp32-p4-86/device/network_coprocessor.yaml
@@ -6,6 +6,7 @@
 # =============================================================================
 
 esp32_hosted:
+  id: esp32_c6_hosted
   variant: esp32c6
   reset_pin: GPIO54
   cmd_pin: GPIO19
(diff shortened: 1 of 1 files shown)


jtenniswood:
Nice idea, I had no idea this was possible. Looks like it can be extended to all supported screens too.

hunter-nl:
> Nice idea, I had no idea this was possible. Looks like it can be extended to all supported screens too.

Yes, that is certainly possible. I see that you already created a PR for it.

PR #584, opened 2026-06-18
Upgrade framework

hunter-nl:
## Summary
During building firmware it shows framework version is not recommended version. Also during compile several warnings shown of issues in the used framework code. These are fixed in newer framework version.
In this PR removed the pinned versions, so it will follow the recommended versions:
Currently recommend versions: esp-idf 5.5.4, pioarduino 55.03.39 with tool-esptoolpy 5.3.0. No issues.

Compliant with ESPHome 2026.6.0

-

## Testing

- [X] I checked the change locally where practical.
- [X] Affected display/device, if applicable:
All


jtenniswood:
Looks good, thanks for that!

Issue #522, opened 2026-06-16
Apply changes live to device


jtenniswood:
See if it's possible to apply changes live to the device and not rely on the apply/reboot to setup

jtenniswood:
# Auto Live Configuration Refresh

## Summary

Make setup changes apply to the panel automatically without rebooting. The first version will use debounced auto-refresh after saved card/layout/config changes, keep the device stable while users are editing, and retain a separate manual restart fallback for recovery only.

## Key Changes

- Add a generated full live-apply script for each device that runs the same grid setup work used at boot:
  - run Phase 1 visual setup
  - wait briefly
  - run Phase 2 Home Assistant subscriptions/subpage setup
  - refresh time/clock bar state afterwards
- Keep the existing lightweight `refresh_button_grid` for layout-only changes where it is already safe.
- Add a shared debounced config-refresh script, using `mode: restart`, so multiple rapid saves collapse into one live refresh.
- Trigger the debounced refresh from:
  - main card config text entities
  - subpage config chunk text entities
  - button order
  - card color settings
  - imported/restored config writes through the same saved entities
- Harden live refresh before enabling repeated rebuilds:
  - delete old dynamically created subpage LVGL screens before clearing navigation registrations
  - close active modals before rebuilding card/subpage UI
  - keep existing Home Assistant subscription generation checks so stale callbacks are ignored
- Update the web UI:
  - remove the current “Apply Configuration” reboot behavior from the primary setup flow
  - show that changes apply automatically
  - keep a clearly separate restart/reboot control only as a recovery action, not as normal setup

## Public Interfaces

- Add internal firmware scripts:
  - `live_config_refresh_debounced`
  - generated per-device full refresh script, for example `apply_live_button_grid_config`
- Keep the existing web API style: the setup page continues posting to ESPHome text/select/number/switch entities.
- Do not change backup/export format.
- Do not require users to reconfigure existing panels.

## Test Plan

- Run generated-file validation:
  - `python3 scripts/generate_device_slots.py --check`
  - `npm run check:generated`
- Run firmware/web checks covering config and UI behavior:
  - `npm run check:firmware-parser`
  - `npm run check:firmware-ha-bindings`
  - `npm run check:firmware-card-runtime`
  - `npm run check:web-smoke`
  - `npm run check:fast`
- Compile at least the affected display firmwares, prioritizing:
  - 7-inch P4
  - 10-inch P4
  - 4-inch S3
- Device test scenarios:
[...]

PR #596, opened 2026-06-19
Fix confirm button theme

jourdant:
## Summary

- This started with the confirmation modal not matching the active button theme: custom button colours worked on the grid, but the final confirm button still used the default slider colour.
- This now carries the selected button’s active colour into the confirmation modal, so the confirm action visually matches the button that opened it.
- Added a readable text colour helper so labels stay legible against both light and dark custom theme colours.
- Applied the contrast handling to themed grid button states, modal selector labels, and confirmation modal buttons.
- Added refresh hooks so configurable theme colour changes redraw the button grid instead of waiting for another refresh.

## Testing

- [x] Affected display/device, if applicable: all devices (I tested against my Guition ESP32-P4 JC8012P4A1)

## Notes for testing

- Test a button with a custom `on` colour and confirm that the confirmation modal’s confirm button uses the same accent colour.
- Test both light and dark custom colours and confirm the button text remains readable.
- Change the configured theme colours and confirm the grid redraws with the updated colours.

## Issue handling

This PR is also loosely related to #522 in the sense that this also adds some live refresh hooks for card colour settings.

### DIFF
--- a/common/config/colors.yaml
+++ b/common/config/colors.yaml
@@ -26,6 +26,8 @@ text:
     update_interval: never
     web_server:
       sorting_group_id: sg_display
+    on_value:
+      - script.execute: refresh_button_grid
 
   - platform: template
     name: "${entity_button_off_color}"
@@ -42,6 +44,8 @@ text:
     update_interval: never
     web_server:
       sorting_group_id: sg_display
+    on_value:
+      - script.execute: refresh_button_grid
 
   - platform: template
     name: "${entity_sensor_card_color}"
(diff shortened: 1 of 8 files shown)


jourdant:
Sorry I just realised I should check if you wanted any further changes before we close the PR

jtenniswood:
Strangely this didn't work for me, but in testing there are also loads of other places where it's not calling the user's primary colour, so will fix those too

jtenniswood:
Nope, all good, appreciate the help!

PR #624, opened 2026-06-20
Hide saved card changes until restart

jtenniswood:
## Purpose
When a user saves a card from the device webserver, the live display can briefly show a card that no longer matches the saved setup. This changes saved card configs so the affected home-screen card is hidden until the display is restarted.

## Practical impact
- Saving a home-screen card marks that card as pending restart.
- The live display hides pending-restart cards instead of showing a stale or misconfigured card.
- On the next restart, firmware clears the pending marker and shows the saved card normally.
- Subpage-only edits are unchanged because they are not standalone home-screen cards.

## Testing
- npm run check:types
- npm run check:web-smoke
- npm run check:generated

## Device testing notes
Flash this PR branch to the affected display, save a home-screen card in the webserver, and confirm the card disappears from the display until the display is restarted. After restart, confirm the saved card appears with the expected setup.


jtenniswood:
Didn't work

PR #691, opened 2026-06-24
Rename cover modal subtype label

jtenniswood:
## Purpose
Rename the Cover card subtype shown as `Modal` to `All Controls` so it matches the wording used for comparable grouped controls.

## Practical impact
In the web setup UI, the Cover card Type dropdown now shows `All Controls` for the existing full-screen cover control option. Existing saved cards still use the same internal `modal` value, so current configurations remain compatible.

The public cover card documentation and generated web UI bundles were updated to match.

## Checks
- `npm run check:web-smoke`
- `npm run check:generated`
- `npm run docs:build`

## Device testing
No firmware flash should be required for this label-only web setup change. To verify on a device, open the setup page, edit a Cover card, and confirm the Type dropdown shows `All Controls` instead of `Modal`.


jtenniswood:
**CI fix in progress** — investigating failing checks on this PR.

Failing checks:
- `Validate Generated Files` — `check_config_formats.js` assertion error: expected `modal:Modal` but got `modal:All Controls`

Diagnosing and will push a fix shortly.

PR #694, opened 2026-06-24
screenshot component for agent interaction

RaymondKroon:
## Summary

I tried to make a virtual device for easy development, however running the webserver is not supported on non ESP devices. So instead I created a screenshot module, which can be used by agents. The agent can also interact with the device on the /touch endpoint.

-

## Testing

- [x] Affected display/device, if applicable:
  Only tested with esp32-p4-86

## Notes for testing

-

## Issue handling


jtenniswood:
Curious about how reliable this is, I see some much variation in devices I don't know that I would trust emulation to be correct.

RaymondKroon:
I wasn't clear. This is not emulating, it is running an additional server on the device. 
You can make a screenshot of what is currently on the display and with the touch endpoint you can interact with the device.

And good point, emulating could be helpful for faster development, but it would not be a real on-device test

jtenniswood:
Thanks, I'm not sure that level of complication is required at this point. Appreciate the offer

PR #706, opened 2026-06-24
Add web UI cleanup

jtenniswood:
## Purpose

Adds a project scan and fixes the safest high-value items from that scan.

## Practical impact

- Reduces fragile dynamic HTML construction in the web UI by using explicit DOM nodes for context menus, icon pickers, card type picker entries, and sun info text.
- Consolidates duplicated POST retry logic in the web UI API helper.
- Rebuilds generated web bundles for all supported layouts.

## Checks

- `npm run check:types`
- `npm run check:generated`
- `npm run check:web-smoke`
- `npm run check:web-browser-smoke`

## Testing notes

This affects the built-in web configuration UI only. Please test opening the panel web UI, editing a card icon/type, using the right-click or long-press context menu, and applying a simple setting change. No firmware flashing is required for this cleanup unless you want to test the generated web bundle on a physical device.


jtenniswood:
Addressed the MDI slug feedback in commit 85e14442d. The shared icon helper now preserves pre-slugged MDI names such as `toggle-switch` and `clipboard-outline`, and the browser smoke test asserts the card type picker keeps `mdi-toggle-switch`.

PR #721, opened 2026-06-26
Subscribe to the entity current_position / percentage / brightness before subscribing to the entity state

kwkoo:
## Summary

- This PR fixes the problem where the slider value is rendered on booting, even if the entity is off.
- In my setup, I have a slider representing a fan entity. When espcontrol boots up, the slider is rendered with the fan speed even if the fan is currently turned off.

## Testing

- [x] User testing is still needed before merge.
- [ ] Affected display/device, if applicable: I have only tested this on an ESP32-P4 86 panel, though the issue should be present across all devices.

## Notes for testing

- Create a slider and bind it to a fan entity
- Turn off the fan
- Reboot the ESP32 panel - the slider should render with a value of 0 because the fan is turned off

## Issue handling


kwkoo:
IMG_3167

Notice how the slider displays the fan speed even though the fan is off. The problem only occurs on the actual display and not in the web interface.

jtenniswood:
Thanks, appreciate the fix

kwkoo:
Thank you!

I tested the main branch and it works.

PR #736, opened 2026-06-26
Conditional colors on sensor card

TomasReznicek:
## Summary

- Add **Conditional Colors** to the Sensor Card: a numeric sensor card recolours its value text based on the live reading crossing user-defined thresholds (e.g. CO2 green → amber → red).
- Each threshold is a `(value, colour)` pair. The colour of the highest threshold at or below the current reading is applied to the value and unit labels on every Home Assistant state update; below the lowest threshold the default tertiary colour is used.
- New `thresholds` sensor option (`kind: color_thresholds`, numeric modes only), stored compactly as `value:RRGGBB,value:RRGGBB`, with a "Conditional Colors" editor (value + colour-picker rows) in the web configurator.
- Two small supporting changes worth calling out for review:
  - Widened the option-`kind` enum in `product_schema.py` and `contracts/types.ts` to allow the new `color_thresholds` kind.
  - Taught `sensor_card_options_normalized()` to preserve `thresholds` through parse-time normalisation (firmware twin of the web normaliser) — without it the option is silently stripped.

## Testing

- [x] User testing is still needed before merge.
- [x] Affected display/device, if applicable: **Guition ESP32-P4 JC8012P4A1**

Local checks all pass (`check_config_formats`, `check_web_smoke`, `check_product_schema`, `tsc`, `build.py --check` zero drift, firmware static checks). Full ESPHome compile (2026.6.1 / ESP-IDF 5.5.4) succeeds for S3-4848 and P4-JC8012P4A1. Verified end-to-end on a physical JC8012P4A1.

## Notes for testing

- Add a numeric Sensor Card → **Conditional Colors** → add rows (e.g. `800` green, `1200` amber, `1600` red) and Save. Value text recolours as the reading crosses thresholds; below the lowest it shows the default tertiary colour.
- Icon/Text mode drops the thresholds (numeric-only); they reappear in Numeric.
- Works identically for sensor cards inside a subpage (shared render path).
- Verified on P4 JC8012P4A1; other panels compile but weren't flashed.

## Issue handling


jtenniswood:
Sorry, I want to keep the presentation simple, I would rather avoid too much customisation.

PR #759, opened 2026-06-27
Reduce date time normalization drift

jtenniswood:
## Summary
- add shared date/time saved-card normalization fixtures for calendar, clock, and timezone cards
- make web editor defaults for date/time cards come from the shared card contract
- align web and firmware cleanup for malformed date/time saved configs
- preserve large-number options only when supported by the shared contract
- refresh generated web bundles for the public device web UIs

## Testing
- python3 scripts/build.py contract --check
- npm run check:config
- npm run check:firmware-card-runtime
- npm run check:product

## Device testing notes
No visible device behavior change is expected for normal saved date/time cards. Calendar cards still use sensor.date, clock cards remain local-only, and timezone cards keep their selected city/timezone while ignored display fields are stripped consistently in web and firmware.


jtenniswood:
Investigating failing CI checks for this PR.

Failing checks:
- **Validate Generated Files**:  failed at 
- Error: 

Working on a fix now...

PR #822, opened 2026-06-29
Lazy build subpage screens

jtenniswood:
## Purpose

Change existing Subpage cards so their hidden screen is built on first open instead of during firmware startup. This is intended as a test branch to judge the real device experience impact before considering nested subpages.

## Practical impact

- Existing saved subpage configuration stays the same.
- Subpage cards still open the same one-level subpage screens.
- The first tap on a subpage may take slightly longer because the screen is created at that moment.
- Later opens reuse the built screen.
- Home Assistant navigation by subpage type can still open a lazy subpage.

## Testing

Automated checks run locally:
- npm run check:fast
- ESPHome compile for `builds/esp32-p4-86.yaml` with local component substitutions

Device testing still needed:
- Confirm the device boots normally.
- Tap existing Subpage cards and compare first-open delay against the current release.
- Open the same Subpage card a second time and confirm it feels normal after the first build.
- Check Back navigation from subpages.
- Check state updates on cards inside subpages after opening them.

## Automated testing guidance

Device testing is expected before merge because this change affects firmware and the on-device experience. Suggested devices to test: ESP32-P4 86 Panel, 7-inch P4, 4.3-inch P4, 10.1-inch P4, and 4-inch S3.

### DIFF
--- a/components/espcontrol/button_grid_actions.h
+++ b/components/espcontrol/button_grid_actions.h
@@ -786,7 +786,7 @@ inline void handle_button_click(const std::string &cfg, int slot_num,
     ha_action_add_data(req, "slot", slot_buf);
     ha_action_send(req);
   } else if (p.type == "subpage") {
-    lv_obj_t *sub_scr = (lv_obj_t *)lv_obj_get_user_data(btn_obj);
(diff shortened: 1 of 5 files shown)


jtenniswood:
Experiment failed. Causes crashing

PR #823, opened 2026-06-29
Guition JC3248W535 3.5" ESP32-S3

lamiskin:
## Purpose

Adds support for the **Guition JC3248W535**, a 3.5-inch 480×320 landscape touchscreen powered by an ESP32-S3.

It's the right size and shape for projects where the other screens are too large.

## What's included

- Full device hardware config (QSPI display via mipi_spi, AXS15231B touch, battery ADC, USB-CDC logging)
- 6 card slots in a 3×2 landscape grid, with rotation support (portrait switches to 2×3)
- Build profiles (standard + factory with bundled web UI)
- Web UI (www.js) with portrait/landscape responsive layout
- Cover art screensaver layout for 480×320
- Manifest, device-profiles, compatibility fixture, and documentation
- README table, install selector, and manual ESPHome setup entries
- S3 memory management (deep-sleep reset workaround, PSRAM WiFi, cover art cleanup)

## How it was tested

- Compiled and flashed over USB to physical hardware
- Verified landscape (3×2) and portrait (2×3) grid layouts on-device
- Confirmed rotation triggers grid recalculation (refresh_button_grid on rotation change)
- Web UI follows rotation state correctly in both orientations
- WiFi, display, touch, and backlight all functional

## Notes

- Shares the same S3 memory constraints as the 4848S040: `imageCardDownloaders: 0`, `apiNavigateAction: false`, `-mtext-section-literals` build flag
- The `portraitCols` / `cols` values in the manifest are intentionally swapped relative to other devices because the native panel is portrait (320×480) with a 90° displayOffset to reach landscape — the generate script's hardcoded 90°/270° = portrait assumption is inverted for this panel
- Hero image pending (placeholder in docs)


jtenniswood:

lamiskin:
Understood.

It would probably be appropriate to spin up "community devices" in to their own repo, so there is a clear distinction of your supported devices, vs community supported ones. I am imagining this new repo would use espcontrol as a git submodule and would only hold the community devices configurations etc (to minimize duplicating your base code).

I am happy to take this on, but where would you like this repo to live? Under your namespace or should it live under an espcontrol org in github? I don't want to limit it to just this one new hardware model, I would like to give others the opportunity to add more devices in the future.

Happy to hear your thoughts to keep the discussion going.

lamiskin:
Circling back to close the loop — I went ahead and built the community devices repo I proposed above: **[lamiskin/espcontrol-community-devices](https://github.com/lamiskin/espcontrol-community-devices)** ([docs & installer](https://lamiskin.github.io/espcontrol-community-devices/)).

I put it under my own namespace for now rather than leave it waiting — happy to move it under an espcontrol org or your namespace later if you'd ever prefer. The final design doesn't use a submodule after all (that fought ESPHome's remote-package model); instead each device pulls your unmodified code at a pinned release, so there's no forked or duplicated base to maintain.

The JC3248W535 from this PR is the first device in it, and it's hardware-verified. Everything's clearly marked unofficial — issues go only to the community tracker, never here, and there's a documented handover path if you ever decide to adopt a device upstream. More context in #1110.

Thanks again for being open to this.

Issue #43, opened 2026-04-16
[Feature Request] Ability to turn screen orientation


nickjayr:
i have the s040 but id like the usb on the left side

jtenniswood:
Might be possible with the new esphome, previously you could only do this at runtime. Will need some investigation.

jtenniswood:
Just added to the latest release

nickjayr:
Thanks!
Le ven. 17 avr. 2026 à 19:02, James Tenniswood ***@***.***> a
écrit :

kevinreid:
I tried rotating 90 and 0 - but the touch mapping doesnt seem to move (ie the image rotated but touch mapping did not).

jtenniswood:
Are you running the latest firmware? I found and fixed a lot of these issues today, so try the latest and let me know.

ouchkilljoys:
Can confirm the display of the image can rotate but the button presses do not, so they do not line up properly with the screen once rotated (version v.1.6.5)

PR #829, opened 2026-06-30
fix: switch non-firmware CI jobs from self-hosted to ubuntu-latest

jtenniswood:
## Problem

The self-hosted GitHub Actions runner has been going offline intermittently, blocking several workflows:

- **CI validate/docs/sync-icons jobs** — These run on `self-hosted` but only need Node.js, Python, and npm, which are all available on standard GitHub-hosted runners.
- **Nightly firmware test build** — 5 compile jobs cancelled today with *"The job was not acquired by Runner of type self-hosted even after multiple attempts"*
- **Release preflight/device-matrix/publish** — These don't need special hardware but are blocked when the runner is down.

Example failing runs: Nightly \#43, CI \#3076

## Changes

| Workflow | Job(s) Changed | Old Runner | New Runner |
|---|---|---|---|
| CI (`ci.yml`) | sync-icons, validate, docs | self-hosted | ubuntu-latest |
| Build Release (`release.yml`) | preflight, device-matrix, publish | self-hosted | ubuntu-latest |

**Intentionally kept on `self-hosted`:** All firmware compile jobs (`nightly-firmware.yml`, `firmware-compile.yml`, `release.yml` build-firmware) and runner disk cleanup — these need Docker + ESPHome + PlatformIO which require the self-hosted runner's hardware.

## Testing notes

- After merge, the next CI run on any PR or push will use `ubuntu-latest` for validate, docs, and sync-icons
- Firmware compilation still requires the self-hosted runner to be online
- A quick way to verify: push a trivial commit to a PR branch and check that the validate/docs jobs run on ubuntu-latest

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
   sync-icons:
     name: Sync Generated Icons
     if: github.event_name != 'pull_request'
-    runs-on: self-hosted
+    runs-on: ubuntu-latest
     permissions:
       contents: write
     outputs:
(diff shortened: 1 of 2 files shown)


jtenniswood:
Reverting — user wants to keep self-hosted runners. Will diagnose the runner offline issue instead.

PR #837, opened 2026-07-01
Fix garage toggle not firing when label display is Status

drnichols:
## Summary
- Toggle-mode garage cards (`garage_mode` empty) with **Label Display** set to **Status** did nothing on tap — `handle_button_click` returned early before reaching `send_toggle_action`.
- Status display is documented (`docs/card-types/garage-doors.md`) as display-only and is already handled separately via `subscribe_garage_state`/`bind_garage_status_card`; it should never gate whether a tap fires a command.
- This predates the change and was found incidentally while building an unrelated new "Gate" card type (fixed separately in that branch) that copied this same dispatch logic. It's a one-line removal of the stray `else if (garage_card_show_status(p)) { return; }` branch.

## Test plan
- [x] `npm run check:firmware-ha-bindings` — passes (including self-test)
- [x] `npm run check:fast` — passes (full suite, exit code 0)
- [ ] **Not verified on physical hardware in this session.** Please flash a device with a garage card configured for Toggle interaction + Label Display: Status and confirm tapping the card now sends the toggle command to Home Assistant as expected. Leaving this PR open until you've had a chance to test on real hardware.


jtenniswood:
Thanks appreciate the fix

PR #858, opened 2026-07-03
Remove moved PR workflow skills

jtenniswood:
## Purpose
Remove the PR review and PR conflict-resolution skills from espcontrol after moving them to espframe.

## Practical impact
- Removes the local `/pr-review` skill copy from espcontrol.
- Removes the local `/resolve-pr-conflicts` skill copy from espcontrol.
- Keeps the source of these skills in espframe instead.

## Checks
- `git diff --cached --check`: passed before commit.

## Device testing
- Not required. This only removes local agent skill instructions.


jtenniswood:
Don't do this

Issue #856, opened 2026-07-03
esphome-intercom crash


hunter-nl:
@jtenniswood Crash loop again with current main branch. :-(

After firmware flashed, it starts and crashed directly. Second start shows Connection lost (check if HA is available). But HA is available, device gets also time from net. So WiFi is working.
But then after few seconds it crashed. And this behavior since second start, keeps repeating.

```
[09:25:35][D][sntp:099]: Synchronized time: 2026-07-03 09:25:35
[09:25:35][I][backlight:096]: Sunrise 05:23, Sunset 22:04 (tz=2.0)
[09:25:41][D][esp32_hosted.update:205]: Fetching manifest
[09:25:41][D][esp32_hosted.update:302]: Best compatible version: 2.12.9
[09:25:41][W][component:513]: esp32_hosted.update took a long time for an operation (351 ms), max is 50 ms
[09:26:02][D][api:228]: Accept 192.168.1.3
[09:26:02][D][api.connection:2461]: Home Assistant 2026.7.0 (192.168.1.3): connected
[09:26:02][I][sensors:2965]: Local sensor endpoint registered
[09:26:02][I][api:022]: API client connected: Home Assistant 2026.7.0
[09:26:02][I][api:039]: Using Home Assistant base URL: http://192.168.1.3:8123
[09:26:02][I][espcontrol:2899]: Local action endpoint registered
[09:26:03][E][esp32.crash:332]: *** CRASH DETECTED ON PREVIOUS BOOT ***
[09:26:03][E][esp32.crash:335]:   Reason: Fault - Load accessGuru Meditation Error: Core  1 panic'ed (Load access fault). Exception was unhandled.
[09:26:03]
[09:26:03]Core  1 register dump:
```

jtenniswood:
Should be fixed now

hunter-nl:
@jtenniswood With main branch of today. It still crash in same behavior.

After power up the device, first crash:
```
[10:16:22][D][i2s_codec_dev:321]: GMF codec audio_stack_tx IO open (io_size=0 buffer_size=0 task_stack=0 prio=0 core=0 psram=no speed=no)
[10:16:22][D][audio_stack:232]:Guru Meditation Error: Core  0 panic'ed (Unknown). Exception was unhandled.
```
The following 9 crashes with same output. (then ESP wil stop the loop permanently until power reset):
```
[...]

PR #859, opened 2026-07-03
Pin ESPHome build version after crash loop

jtenniswood:
## Purpose

Issue #856 reports a crash loop on firmware built from current `main`. The only build-environment change immediately before the report was the automated firmware workflows moving from ESPHome `2026.6.2` to `2026.6.4`.

This PR pins the automated firmware build version back to ESPHome `2026.6.2`, the previous version used by the release, nightly, and compile workflows.

## Practical impact

New firmware built by GitHub Actions should avoid the ESPHome `2026.6.4` build path while the crash-loop report is investigated. No device UI, card behavior, or stored configuration format is changed.

Related to #856. The issue should remain open until the affected display has been flashed and confirmed stable.

## Checks

Automated checks run locally:
- `npm run check:fast`

Device testing:
- Not yet tested on physical hardware.
- Please flash the affected display from this PR build and confirm it boots, connects to Home Assistant, and remains stable past the previous crash point.


jtenniswood:
Fixing issues with 2026.6.4 instead

hunter-nl:
@jtenniswood The crash is not related to esphome version. It is related to audio stack implementation.

PR #1011, opened 2026-07-11
Improve P4-86 dual mic audio stability

salnajjar:
This PR updates the ESP32-P4 86 audio setup after testing the Waveshare ESP32-P4-WIFI6-Touch-LCD-4B board with the ES7210/ES8311 audio path.

The board exposes the ES7210 input as a four-slot TDM stream: mic 1 on slot 0, the hardware speaker reference on slot 1, mic 2 on slot 2, and slot 3 unused. The old config was only feeding one mic slot into the stack, so the AFE was not getting the layout it needs for dual-mic speech enhancement and AEC.

Related to #856.

What changed:

- Set the AFE to `type: sr`, `mode: high_perf`, `mic_num: 2`, `input_format: MMNR`, and `se_enabled: true`.
- Mapped the ES7210 TDM mic slots to `[0, 2]`, with the hardware reference on slot `1`.
- Kept AEC enabled, with the non-linear processing level set to `normal`.
- Disabled AFE noise suppression and AGC so the dual-mic speech enhancement path is doing the main clean-up work.
- Set `memory_alloc_mode: more_psram`.
- Kept the AFE feed and fetch frame buffers in internal RAM to reduce latency and improve responsiveness.
- Left the AFE feed ring in PSRAM. Setting `feed_ring_in_psram: false` caused a memory-pressure segfault, so it cannot safely be moved internal on this build.
- Tuned the ES7210 microphone gain to `36.0` after repeated voice and playback testing.
- Kept the AFE input/output diagnostic sensors in the firmware, but made them internal and returned them to a 500 ms update interval so they do not publish noisy entities back to Home Assistant.
- Reduced the hosted WiFi/LWIP buffer pressure for this device and disabled high-performance networking, leaving more internal/DMA-capable memory available while audio and the P4/C6 hosted WiFi link are both busy.
- Added `dev-docs/p4-86-audio-stack-test.md` with the testing notes and the reasoning behind the changes.

Why:

During testing, dual-mic speech enhancement and playback could leave too little DMA-capable memory for the hosted WiFi/SDIO receive path. That showed up as `sdio_rx_get_buffer` asserts and reboots. The audio changes give the AFE the correct mic/reference layout, and the networking changes reduce burst pressure so voice processing and playback have a better chance of co-existing cleanly.

The latest buffer placement keeps the time-sensitive AFE frame buffers internal for better response, while leaving the feed ring in PSRAM because putting everything internal pushed the board into a segfault.

The microphone gain was then raised in small steps while testing real voice commands. `36.0` is the current test value: it improves pickup without adding extra Home Assistant diagnostic chatter.

Testing:

- Built and flashed `devices/esp32-p4-86/dev.yaml` over USB on `COM3`.
- Re-tested after restoring normal cover art downloads, the 20 MHz SDIO clock, 500 ms speaker/mixer/resampler buffers, and `media_player.buffer_size: 1000000`.
- Latest flashed build uses `mode: high_perf`, `input_format: MMNR`, `aec_nlp_level: normal`, `memory_alloc_mode: more_psram`, `feed_buf_in_psram: false`, `feed_ring_in_psram: true`, `fetch_ring_in_psram: false`, and ES7210 `gain_db: 36.0`.
- Tested on the P4-86 panel with voice commands, longer TTS responses, and the media playback path that previously triggered crashes.
- Tried `feed_ring_in_psram: false`; it caused a segfault due to lack of internal memory, so that setting was reverted before updating this PR.


salnajjar:
@jtenniswood I've asked @hunter-nl to test this PR before putting as ready for merging.

salnajjar:
@jtenniswood This is good to be merged into main when you're able. @hunter-nl has confirmed alongside myself that it's functioning nicely :)

Issue #1002, opened 2026-07-11
02 — Internal: Dependency-Aware Check Graph


jtenniswood:
# Dependency-Aware Check Graph

## Summary

Replace the long aggregate check commands with a small Python task orchestrator, delivered through six sequential PRs. Existing validators and generators remain unchanged. Version one stays sequential and fail-fast; later PRs add changed-file routing, opt-in parallelism, and local-only caching.

Each PR will use a new worktree and short-lived branch based on the latest `main`, be committed and pushed, and be opened ready for review. Each subsequent PR starts only after the previous one is merged.

## Interfaces and data model

Add a pure task registry plus CLI runner:

- `scripts/check_tasks_data.py`: task definitions with ID, command arguments, dependencies, profiles, domains, input globs, generated inputs, and cache policy.
- `scripts/check_tasks.py`: planning, execution, summaries, changed-path selection, parallel scheduling, and caching.
- Keep existing `npm run check:*` names as compatibility entry points through `package.json`.

Supported commands:

```text
python3 scripts/check_tasks.py list [--json]
python3 scripts/check_tasks.py plan <profile> [--domain <domain>] [--json] [--explain]
python3 scripts/check_tasks.py run <profile> [--domain <domain>] [--jobs N] [--no-cache] [--summary-json PATH]
python3 scripts/check_tasks.py run-task <task-id>
python3 scripts/check_tasks.py changed [--base REF] [--explain] [--json]
python3 scripts/check_tasks.py cache status|clear
python3 scripts/check_tasks.py --self-test
```

Profiles are assurance levels: `product`, `fast`, `ci`, `all`, and `release`. Domains are optional filters: `product`, `firmware`, `web`, `docs`, and `workflow`. Dependency closure is always added after filtering.

JSON summaries use a versioned schema containing profile, domain, Git revision, timing, cache state, and per-task `passed`, `failed`, `blocked`, `not_run`, or `cached` status. Exit codes are `0` for success, `1` for check failure, `2` for invalid configuration/usage, and `130` for interruption.

### PR 1 — Define declarative check task registry

Branch: `define-check-task-registry`  
Title: `Define declarative check task registry`

- Record every existing focused check and aggregate membership without changing execution.
- Model generated-output validation as a read-only prerequisite for checks that consume generated files.
- Use stable topological ordering, with declaration order as the tie-breaker.
[...]

PR #1021, opened 2026-07-12
Add opt-in parallel check execution

jtenniswood:
## Purpose

Deliver stage 5 of #1002 by adding safe, opt-in parallel execution to the dependency-aware check graph.

## Practical impact

- Adds `--jobs N` while keeping every existing npm alias and CI on the sequential one-worker default.
- Adds `npm run check:parallel`, capped at four workers for local fast checks.
- Runs only dependency-independent tasks explicitly declared parallel-safe at the same time.
- Keeps browser, release, Git-state, and shared-output tasks exclusive; release profiles always use one worker.
- Captures concurrent output and replays complete task blocks in stable order.
- Stops scheduling after the first failure, lets active tasks finish, and reports dependent tasks as blocked.
- Records worker use in human, JSON, and GitHub summaries and exposes safety policy in the generated check matrix.

## Automated checks

- `python3 scripts/check_tasks.py --self-test`
- `npm run check:parallel`
- `npm run check:ci` (including browser smoke)
- `python3 scripts/check_dev_docs.py --check`
- `npm run docs:build`

The self-test covers dependency-safe concurrency, grouped output, exclusive unsafe tasks, fail-stop scheduling, dependency blocking, active-task completion, release serialization, invalid worker counts, missing executables, and interruption of every active child process.

## Repeatability and performance

On Darwin arm64, 20 sequential and 20 four-worker complete fast runs all passed with identical task-status results and an unchanged tracked-diff fingerprint. Median runtime improved from 11.850s to 9.635s (18.7%). Because this is below the planned 20% threshold, parallel execution remains explicit-only and all normal aliases and CI remain sequential.

## Testing notes

No firmware or end-user device behaviour changes, so physical device testing is not required. To test locally, run `npm run check:parallel` and confirm output appears in complete per-task blocks followed by a passing summary showing four workers.


jtenniswood:
@codex review

PR #1095, opened 2026-07-14
Screen Lock: configurable lock display and unlock methods

CaputoDavide93:
# Screen Lock: configurable lock display and unlock methods

## Purpose

Today the Screen Lock card is a single-tap toggle that guards other cards locally. This PR turns it into a configurable lock while keeping the default behaviour identical for existing cards.

Two new dropdowns and a PIN field on the Screen Lock card:

- **When Locked** — `Screensaver` (default) or `Screen Off`: what the panel shows once locked.
- **Unlock With** — `Immediate` (default, single tap), `Slide` (drag a handle bottom→top, like the light-brightness slider), or `PIN` (numeric keypad).
- **PIN** — digits, shown only when Unlock With = PIN. Blank falls back to immediate so a user can never lock themselves out.

It also exposes the lock to Home Assistant for the first time: a **Screen Lock** switch (lock/unlock remotely) and a **Screen Locked** binary_sensor, kept in sync when the panel is locked/unlocked on-device.

## Practical impact

- A panel in a child's/shared room can be genuinely locked, dark (or on the screensaver), and require a deliberate slide or a PIN to reopen.
- Existing Screen Lock cards keep working unchanged: default is Screensaver + Immediate, which matches today's behaviour aside from also publishing the new HA entities.
- The PIN is stored locally in the card config (visible via the panel web API); it is a deterrent, not strong security. Documented as such.

## How it works (follows the card contract)

- **Contract** (`common/config/card_contract.json`): three declarative options on `screen_lock`, stored in spare base fields (`sensor` = lock display, `unit` = unlock method, `entity` = PIN), mirroring how `internal` reuses `sensor`. Generated TS/firmware/docs regenerated via `build.py`.
- **Web configurator** (`src/webserver/cards/screen_lock.ts`): two segment controls + a numeric PIN field. `config_codec.ts` now keeps those base fields for `screen_lock` (they were previously stripped).
- **Firmware**: options captured at card registration (`button_grid_cards.h`); on lock, a display callback drives the screensaver or backlight-off; unlocking is gated by method in a shared `screen_lock_handle_tap` used by both the main-grid and subpage tap paths. New `components/espcontrol/button_grid_screen_lock.h` holds the slide-to-unlock modal and the PIN keypad, which **reuses the existing alarm keypad** building blocks. Two new `ControlModalKind` values.
- **Home Assistant**: non-internal `switch` + `binary_sensor` and a publish callback in `common/config/display.yaml` so on-device unlocks stay in sync (no feedback loop). These are the first non-`internal` entities in the project.

## How it was checked

- `npm run check:ci` passes end to end (contract/model/backup, saved-config parity browser↔firmware, saved-config production, `firmware-tests` cmake/ctest 4/4, the firmware static analyzers, web + browser smoke across all 6 device layouts, release checks, docs build).
- `python3 scripts/build.py --check` clean; `tsc --noEmit` clean.
- A full ESP-IDF firmware compile for `guition-esp32-s3-4848s040` succeeds (`esphome compile` → "Successfully compiled program", firmware.bin created; RAM 37.7%, Flash 43.9%).

## Testing notes (please test the branch, not main)

**A build/analyzer pass is not a device test.** The slide-to-unlock modal and the PIN keypad are LVGL UI that should be confirmed on hardware.

Affected device for flashing: **Guition ESP32-S3-4848S040** (and any device that includes the shared `screen_lock` card + `display.yaml`).

Suggested manual checks after flashing:

1. Add a Screen Lock card; set When Locked = Screensaver, Unlock With = Immediate → lock drops to screensaver, a tap unlocks (unchanged default).
2. Switch to Screen Off → locking turns the backlight off; the configured unlock still works.
3. Unlock With = Slide → locked screen shows the slide control; dragging to the top unlocks, releasing early stays locked.
4. Unlock With = PIN, set a PIN → keypad appears; correct PIN unlocks, wrong PIN clears + flashes; clearing the PIN in config falls back to immediate.
5. Home Assistant: the `Screen Lock` switch and `Screen Locked` sensor track on-device locking/unlocking, and toggling the switch locks/unlocks the panel.

## Notes for review

- Introduces the project's first non-`internal` (HA-exposed) entities — flagged in case you'd prefer a different naming/entity_category convention.
- PIN is stored in plaintext in the compact config string; happy to gate it differently if you'd rather.


CaputoDavide93:
Follow-up: ran an adversarial review pass over the firmware and tightened several edge cases (branch updated, still compiles for ESP32-S3):

- **Screen-off + open control modal:** closing the panel while a climate/light/cover/media modal is open now force-closes that modal first (`control_modal_force_close_active()` in the display callback) so it can't stay operable on the top layer or cover the unlock card while locked.
- **Unlock modal lifetime:** a remote (HA) unlock while the slide/PIN modal is open now dismisses the modal, and a config/registry rebuild tears it down first (`screen_lock_close_unlock_modal()`), avoiding a modal surviving over an unlocked/rebuilt grid and dangling static UI pointers.
- **PIN validation:** a stored PIN that the keypad can't reproduce (non-numeric or >16 chars) now falls back to immediate unlock instead of trapping the user.
- **Docs:** noted that the Screen Lock card should live on the home screen (a subpage-only lock card can only be unlocked from HA once locked), and that multiple lock cards share one configuration.

A subpage-only lock card remaining unreachable once locked is documented rather than code-fixed for now — happy to add a code guard if you'd prefer.

CaputoDavide93:
Withdrawing for now — device testing on the 480x480 panel showed the slide-to-unlock is still too tall and inconsistent (occasionally unlocks then immediately re-locks). Needs more UX work (likely making PIN/immediate the recommended unlock on small square screens, and reworking the slide control). Will revisit after further on-device testing.

PR #1164, opened 2026-07-16
Suggest Home Assistant entities in entity boxes

RyanEwen:
## Summary

- Entity fields on the setup page autocomplete against the user's real Home Assistant entities: the page connects to HA's WebSocket API using an address and long-lived access token stored on the device (two new settings under **Home Assistant Settings**) and feeds the entity list into the shared suggestion store.
- Selecting a suggestion auto-fills the card label from the entity's friendly name; bare addresses default to port 8123; manual free-text entries are no longer persisted as suggestions.
- Focusing an entity box re-fetches the list when stale, so entities newly created in Home Assistant appear without reloading the page.
- The address and token round-trip through backup export/import.

## Practical impact

- Configuring cards no longer requires typing full entity IDs from memory — pick from a live list of your actual entities, with labels filled in automatically.
- Entirely optional: without an address/token the entity boxes behave exactly as before.

## Documentation decision

- [ ] Updated public docs or release-facing notes for user-visible behavior/configuration.
- [x] Docs follow-up needed before merge; explain below.

Docs notes:

- The setup docs should mention the new Home Assistant address/token settings and how to create a long-lived access token; happy to add in this PR if preferred.

## Testing

- Web setup page for all panels (device side only adds the two settings entities). Tested with a Guition JC4880P443 (4.3") against a live Home Assistant instance.

## Notes for testing

- With address + token set: entity boxes suggest live HA entities; selecting one fills the label; entities created in HA afterwards appear after refocusing the field.
- Without them: no behavior change; the status line under the settings explains how to enable suggestions.
- Token is stored on the panel like other settings; worth a reviewer's eye on whether that warrants a docs caveat.

## Automated testing guidance

Device testing is expected before merge because this change may affect firmware or the on-device experience.

Suggested devices to test:
- ESP32-P4 86 Panel (4 inches, Square)
- Guition JC1060P470 (7 inches, Landscape)
- Guition JC4880P443 (4.3 inches, Portrait)
- Guition JC8012P4A1 (10.1 inches, Landscape)
- Guition JC8012P4A1 new panel (2624+) (10.1 inches, Landscape)
- Guition 4848S040 (4 inches, Square)

Suggested checks:
- Confirm the device boots normally and does not restart repeatedly.
- Confirm the affected screen, card, or setting appears correctly.
- Confirm touch/navigation still works where the changed area is interactive.
- Confirm there is no obvious text overlap, clipping, blank screen, or missing icon.

Suggested PR status: Ready for device test once automated checks pass.

Changed files considered:
- `common/config/display.yaml`
- `common/config/entity_names.json`
- `common/config/entity_names.yaml`
- `docs/public/webserver/www.js`
- `product/product_snapshot.json`
- `src/webserver/application/app_backup.ts`
- `src/webserver/application/app_state_event_handlers.ts`
- `src/webserver/application/entity_state.ts`
- `src/webserver/application/ha_entity_fetch.ts`
- `src/webserver/application/settings_system_section.ts`
- `src/webserver/entry.ts`
- `src/webserver/generated/entity_catalog.ts`
- `src/webserver/model/settings.ts`
- `src/webserver/runtime/application_globals.d.ts`
- `src/webserver/state/app_state.ts`
- `src/webserver/state/types.ts`

## PR status

- [x] Ready for device test.
- [x] Device tested successfully. (JC4880P443 + live HA; broader panel coverage pending)

## Issue handling


jtenniswood:
Interesting, but asking users to setup tokens isn't a smooth user experience.

RyanEwen:
> Interesting, but asking users to setup tokens isn't a smooth user experience.

Counter: Having to lookup and then copy/paste entity IDs and names is not a smooth user experience. Also I found the autocompletes were filling with partial / wrong suggestions over time as well.

jtenniswood:
Auto completes will just be entities you are already using, so they shouldn't be wrong.

RyanEwen:
Might be a bug then

I was seeing `lights.be`, `lights.bedroo` etc as well as the actual `lights.bedroom_lights` option.

Regardless, I found it unhelpful to only be able to autocomplete the entities that I already added, vs the ones I am wanting to add.

mike391:
> Auto completes will just be entities you are already using, so they shouldn't be wrong.

Why the resistance? I think this would be a good QOL improvement as well

PR #1208, opened 2026-07-18
Notifications

wonderslug:
# HA-driven notification popups: text + image, with accent-tinted image sizing

## Summary

Adds a top-layer LVGL overlay that Home Assistant can push notifications to via new
ESPHome native API actions — for both **plain text** and **images** (e.g. a doorbell
camera snapshot). The popup renders above every page (main grid, subpages, clock
screensaver), wakes the screensaver on arrival, and blocks interaction with the UI
beneath. On dismiss — by tap or by timeout — the device fires an event back to HA
carrying a `message_id` so automations can tell a *read* notification from an
*unread/expired* one.

The device only ever receives a **URL** for images; Home Assistant does all image
sourcing (camera proxy, `/local/…`, or any `http(s)` URL). The device downloads and
decodes (JPEG/PNG) on-panel, capping the requested source resolution and checking free
heap before starting the download. If the fetch or decode fails — or memory is too
tight to safely attempt it — the popup **falls back to a text notification** so the
alert still gets through.

Image notifications render in the same inset rounded panel, top-right X close button,
and shadow-text title/message overlaid directly on the image that every other control
screen in the app already uses, so they look consistent with the rest of the UI instead
of introducing a one-off layout.

---

## Screenshots

**Text notifications**

IMG_2253

**Image notification — with cover / crop to fit**
IMG_2258

**Image notification — with fit / full image**
IMG_2257

---

## What's new vs. #134

| | #134 | This PR |
|---|---|---|
| Text notifications | ✅ | ✅ (unchanged) |
| Image notifications | — | ✅ `send_image_notification` / `send_expiring_image_notification` |
| Image sizing modes | — | ✅ `cover` / `fit` |
| Accent-tinted letterbox | — | ✅ padding sampled from the image itself |
| P4 hardware acceleration | — | ✅ hardware JPEG decode + background download on ESP32-P4 panels |
| Text ↔ image fallback | — | ✅ failed/oversized download → text popup |
| Shared close events | ✅ | ✅ text **and** image fire the same events |
| Base branch | older `main` | rebased onto current `main` |

---

## New API actions

All actions are exposed per-device as `esphome.<device_name>_<action>`. ESPHome API
action variables have **no defaults** — every declared variable is required from HA's
perspective — so optional-feeling fields (`title`, `message`, `message_id`,
`image_sizing`) are passed as empty strings when unused, and the firmware applies the
documented default.

### `send_notification` — persistent text (tap to dismiss)

| Variable | Type | Required | Description |
|---|---|---|---|
| `message` | string | yes | Body text. |
| `title` | string | optional (`""`) | Heading; hidden when empty. |
| `message_id` | string | optional (`""`) | Correlation id echoed back in close events. |

### `send_expiring_notification` — auto-dismissing text

Same as above **plus**:

| Variable | Type | Required | Description |
|---|---|---|---|
| `timeout` | int | yes | Seconds until auto-dismiss. |

### `send_image_notification` — persistent image (tap to dismiss)

| Variable | Type | Required | Description |
|---|---|---|---|
| `image_url` | string | yes | Relative HA path (e.g. `/api/camera_proxy/camera.front_door`) or any `http(s)` URL. |
| `title` | string | optional (`""`) | Caption heading. |
| `message` | string | optional (`""`) | Caption body. |
| `image_sizing` | string | optional (`""` → `cover`) | `cover` or `fit`. |
| `message_id` | string | optional (`""`) | Correlation id echoed in close events. |

### `send_expiring_image_notification` — auto-dismissing image

Same as `send_image_notification` **plus** `timeout` (int, seconds, required).

#### Example (bare action call)

```yaml
action: esphome.living_room_panel_send_expiring_image_notification
data:
  image_url: "/api/camera_proxy/camera.front_door"
  title: "Front Door"
  message: "Motion detected"
  image_sizing: "cover"
  timeout: 20
  message_id: "doorbell"
```

#### Example (full automation — doorbell snapshot to the panel)

```yaml
automation:
  - alias: "Doorbell → panel image notification"
    trigger:
      - platform: state
        entity_id: binary_sensor.front_door_button
        to: "on"
    action:
      - action: esphome.living_room_panel_send_expiring_image_notification
        data:
          image_url: "/api/camera_proxy/camera.front_door"
          title: "Front Door"
          message: "Someone's at the door"
          image_sizing: "cover"
          timeout: 30
          message_id: "front_doorbell"
```

---

## Image sizing modes

The image is decoded into a panel sized and positioned the same way as every other
control screen's panel — computed at runtime from the device's screen dimensions, not
a fixed per-device constant — then shown in a rounded-corner container (clipped via an
LVGL 9 `clip_corner` frame).

| `image_sizing` | Resize mode | Result |
|---|---|---|
| `cover` *(default)* | `COVER` | Crops to fill the panel edge-to-edge, no bars. |
| `fit` | `FIT` | Shows the **whole** image; leftover letterbox padding is tinted with an accent color sampled from the image (see below), instead of black bars. |

### Source resolution capping + memory guard

Before downloading, the requested source resolution is capped to 800px on its long
side (for recognized HA camera/image proxy URLs, via a `?width=&height=` query param)
so the JPEG decode workspace scales with a sane cap instead of an arbitrary source
image's full resolution — the same capping the tap-to-expand camera card modal already
uses. A pre-flight free-heap check runs first; if memory is too tight to safely attempt
the download, the notification falls straight back to the text popup instead of
risking an oversized decode.

On ESP32-P4 panels, the download itself runs on a background task with hardware JPEG
decode + scaling, the same acceleration tier the camera card's tap-to-expand modal
uses. On ESP32-S3 panels these settings are inert no-ops — behavior is unchanged.

---

## HA events fired on close

Both text and image notifications fire the **same two events**, so one HA automation
handles either kind:

| Event | When | Payload |
|---|---|---|
| `esphome.notification_acknowledged` | User tapped to dismiss | `device_name`, `message_id` |
| `esphome.notification_expired` | Auto-timeout elapsed unread | `device_name`, `message_id` |

This lets automations distinguish "user saw it" from "it timed out unread" — useful for
critical alerts that should escalate if ignored.

---

## Design decisions

- **Two actions instead of one optional `timeout`.** ESPHome action variables have no
  defaults, so a single action with a `timeout` field would always require it. Splitting
  into persistent vs. expiring makes intent explicit at the call site. (Carried over
  from #134; applied to the image actions too.)
- **Top-layer overlay, not a new LVGL page.** Rendering on `top_layer` places the popup
  above the main grid, subpages, and the clock screensaver with no page navigation. The
  screensaver is woken before showing.
- **Touch blocking scrim.** The full-screen scrim is `clickable: true` and consumes all
  touches, preventing accidental button presses while a notification is up. Tapping
  anywhere dismisses.
- **Distinct acknowledged vs. expired events** (see above).
- **Device receives only a URL.** HA does all image sourcing; the panel downloads and
  decodes. Keeps the firmware agnostic to camera/entity plumbing.
- **Graceful fallback.** A failed download/decode, or a pre-flight memory check that
  fails, falls back to the text popup with the same `title`/`message`/`timeout`/
  `message_id` (or "Image unavailable" if image-only), so an alert is never silently
  dropped.
- **Reuses the existing cover-art accent sampler.** `fit`-mode letterbox tinting calls
  the same `extract_accent_color_rgb565` / `darken_accent_color` helpers already used by
  cover art — no duplicate color logic, no changes to `cover_art.h` or
  `screen_cover_art.yaml`.
- **Reuses the camera card's panel geometry, sizing cap, and memory guard.** Rather than
  a separate layout or sizing scheme, the popup calls the same layout, source-size-cap,
  and free-heap-check helpers the camera card's tap-to-expand image view already uses —
  one tested code path for both, and the P4 acceleration settings apply the same way in
  both places.

---

## Files

Shipped as common packages included by all devices:

| File | Role |
|---|---|
| `common/device/screen_notification_popup.yaml` | Text popup: LVGL widget + `send_notification` / `send_expiring_notification` actions |
| `common/device/screen_image_notification.yaml` | Image popup: LVGL widgets + image actions + sizing/tint logic |
| `components/espcontrol/notification_popup.h` | Text: C++ show/hide, auto-dismiss timer, event firing |
| `components/espcontrol/image_notification_popup.h` | Image: state + event helper (reuses the text popup's event payload) + URL resolver |
| `components/artwork_image/artwork_image.{h,cpp}` | Adds `paint_padding()` for accent-tinted letterbox bars |
| `components/espcontrol/button_grid_image.h` | Camera card image helpers reused as-is: panel layout, source-size capping, memory guard |
| `components/espcontrol/button_grid_modal.h` | Shared panel/chrome layout (`control_modal_calc_layout`, close-button styling) reused as-is |

> Image notifications depend on the text popup: `image_notification_popup.h` includes
> `notification_popup.h` for `notification_fire_event`, and the image download-failure
> path calls `notification_popup_show` for its text fallback — the two ship together.
> `image_notification_popup.h` also includes `button_grid.h` to reach the camera card's
> layout/sizing/memory-guard helpers, all reused verbatim from a new call site.

---

### `notification_popup.h`

```cpp
// Fire a HA event carrying { device_name, message_id }.
void notification_fire_event(const char *event_name, const std::string &message_id);

// Show a text popup. Call after screensaver_wake so the display is on.
// scrim/title_lbl/body_lbl : LVGL objects for the overlay, heading, body.
// title  : heading text; hidden when empty.
// body   : body text.
// timeout_sec : 0 => manual dismiss only; >0 => auto-dismiss after N seconds.
// message_id  : echoed back in the close event.
void notification_popup_show(lv_obj_t *scrim, lv_obj_t *title_lbl, lv_obj_t *body_lbl,
                             const std::string &title, const std::string &body,
                             int timeout_sec, const std::string &message_id);

// Dismiss the active popup.
// user_tapped = true  -> fires esphome.notification_acknowledged
// user_tapped = false -> fires esphome.notification_expired
void notification_popup_dismiss(bool user_tapped);
```

Internally, `notification_timer_cb` is the LVGL timer callback that hides the scrim and
fires `notification_expired` when a `send_expiring_notification` timeout elapses unread.

### `image_notification_popup.h`

```cpp
// Record the message_id of the image notification currently on screen.
void image_notification_set_msg_id(const std::string &message_id);

// Fire a HA event (acknowledged/expired) with device_name + the stored message_id,
// reusing notification_fire_event so text and image events are byte-identical to HA.
void image_notification_fire(const char *event_name);

// Resolve image_url against the cover-art HA base URL.
//   http(s)://... -> returned unchanged
//   relative path -> appended to ha_base_url (slash-normalized)
// Pure string work; no artwork_image dependency.
std::string image_notification_resolve_url(const std::string &image_url,
                                           const std::string &ha_base_url);
```

### `artwork_image` (new method)

```cpp
// Paint the unused padding pixels (outside the decoded content rect — e.g. FIT-mode
// letterbox bars) with a solid RGB565 color instead of the decode buffer's default
// black. No-op when there is no active buffer or no padding. RGB565 images only.
void ArtworkImage::paint_padding(uint8_t r, uint8_t g, uint8_t b);
```

### Reused cover-art accent helpers (`components/espcontrol/cover_art.h`)

`fit`-mode letterbox tinting reuses the existing sampler — no changes to these:

```cpp
struct AccentColor { uint8_t red{0}, green{0}, blue{0}; bool valid{false}; };

// Weighted-average accent color over the decoded content rect (vivid pixels weigh
// more). content_* out of range falls back to sampling the whole image.
// Returns {valid=false} if data is null / dimensions are non-positive.
AccentColor extract_accent_color_rgb565(const uint8_t *data,
                                        int image_width, int image_height, bool big_endian,
                                        int content_x, int content_y,
                                        int content_width, int content_height);

// Darken an accent (each channel / 3) for use as a subdued backdrop/letterbox tint.
AccentColor darken_accent_color(AccentColor color);
```

The image popup samples the decoded image, darkens it, and passes the result to
`paint_padding` so the letterbox bars pick up the image's dominant color.

### Reused camera-card image helpers (`components/espcontrol/button_grid_image.h`)

Called with `ctx = nullptr`, which these functions already treat as "skip the optional
per-entity diagnostic logging" — no changes needed to any of them:

```cpp
// Cap width/height to IMAGE_CARD_MODAL_MAX_TARGET_SIDE_PX (800) on the long side,
// preserving aspect ratio.
void image_card_limit_target_size(lv_coord_t source_width, lv_coord_t source_height,
                                  int *target_width, int *target_height);

// Append ?width=&height= to recognized HA camera/image proxy URLs; any other URL
// passes through unchanged.
std::string image_card_sized_url(const std::string &url, lv_coord_t width, lv_coord_t height);

// Pre-flight check: is there enough free/contiguous heap for a decode of this size?
// Logs a warning and returns false (skip) if not; ctx may be nullptr.
bool image_card_memory_available(ImageCardCtx *ctx, const char *stage, int width, int height);
```

### Reused panel layout (`components/espcontrol/button_grid_modal.h`)

```cpp
// Computes the panel's position/size/corner-radius/inset for the current screen,
// the same geometry every other control screen's panel uses.
ControlModalLayout control_modal_calc_layout(int width_compensation_percent);
```

---

## Testing

Full ESP-IDF compile (ESPHome 2026.7.0) against local components — **all notification
devices pass**:

| Device | Result |
|---|---|
| esp32-p4-86 | ✅ Successfully compiled |
| guition-esp32-p4-jc1060p470 | ✅ Successfully compiled |
| guition-esp32-p4-jc4880p443 | ✅ Successfully compiled |
| guition-esp32-p4-jc8012p4a1 | ✅ Successfully compiled |
| guition-esp32-p4-jc8012p4a1-v2 | ✅ Successfully compiled |
| guition-esp32-s3-4848s040 | ✅ Successfully compiled |

CI (`Validate Generated Files` — incl. `device-slots` / `device-profiles` / `product-snapshot`)
passes on the generated device wiring.

Manual on-device checks: text persistent/expiring, image `cover`/`fit` sizing, relative
+ absolute URLs, download-failure and low-memory fallback to text, and both close
events reaching HA with the correct `message_id`.


RyanEwen:
I love this idea

JLkiller007:
I also love this idea!

jtenniswood:
Nice. 
Couple of thoughts...
- I think this should use the modal pattern (full screen card rather than partial overlay)
- I would add a X close control for the top right
- Rather than use a box overlay, I would use the same text effect used on the camera card (duplicate text in dark colour, 1px offset, with some transparency). I think this is readable against most content.
- Change the image to fill the space, like the camera card modal view.

wonderslug:
This kinda what you had in mind?

full image
IMG_2257

crop to fit
IMG_2258


Are you thinking about not having the original "full screen" and card at all with only this mode?



I kinda feel the text can get lost without something to offset it,  see the P4 crop to fit.

jtenniswood:
Yes that looks great

jtenniswood:
Yes I was thinking just the full card rather than the 'screen saver' style full screen take over.

jtenniswood:
Is this using the p4 hardware acceleration when possible (eg supported hardware). I did more improvements to the camera card/full modal that makes it a lot faster.

wonderslug:
@jtenniswood I updated the branch and force pushed so any checkouts you have need to be re-pulled. This should be a clean, rebased on latest main branch.  It has the modal style only, with the two image sizing options.  

It is now using the updated p4 specific optimizations for speed, and url size limiting for the camera images.

wonderslug:
Still seems to be ready the merge cleanly.  Is there something that needs to be looked at?  

Been running stable locally for a couple of months.  Even use it to send random targeted memes to my son to tell him to do stuff and get off the video games.

wonderslug:
Cleaned up merge to make clean CI commit

PR #1246, opened 2026-07-20
Add a photo screensaver and a calendar agenda card

RyanEwen:
## Photo screensaver + calendar agenda card

Two related features that lean on the existing app plumbing. A **photo screensaver** that cycles images with optional clock / weather / agenda overlays, and a **calendar agenda card** (plus a full-screen agenda view) that pulls events straight from Home Assistant.

| Screensaver | Agenda card | Full-screen view |
|---|---|---|
|  |  |  |

### Photo screensaver
- Cycles images from a file source when the display idles.
- Optional overlays: clock (top-left), weather (top-right), and a compact agenda (bottom). Each toggles independently from the web UI.
- Tapping the agenda overlay opens the full-screen agenda view.

### Agenda card
- A grid card that groups upcoming events by day, with a today marker and week separators.
- **Per-calendar colours** — pick a colour per calendar in the configurator; the card and overlays use it as the event accent.
- Per-card **Days Ahead** limit and an optional **Hide Empty Days**.
- Tap the card to open the same full-screen agenda view.

### Data path
Events come from Home Assistant's `calendar.get_events` via the action-response JSON mechanism (the same path the weather forecast card already uses). A response template renders events into a compact array device-side, so the firmware receives a small payload it parses into a time-ordered, day-grouped list. The parsing/grouping/colour logic is host-tested.

### Web configurator
New settings sections for the screensaver (photo source, weather + agenda overlays, days/limit/hide-empty, date toggle) and for the agenda card (calendar picker with colour selection, label, days ahead, hide empty days). Conditional fields are grouped under an accent rail so dependent options are visually nested.

### Notes
- New host tests: `tests/firmware/calendar_agenda_test.cpp`, `tests/firmware/photo_screensaver_test.cpp`.
- Built on top of the central-app-owner refactor already on main (`espcontrol_app.display()`), not the old global controller.
- Rebased onto current `main`; all `check_tasks` targets pass.
- ~110 files, +7.4k/-99 (new `components/espcontrol/*` headers, web configurator, and device YAML).


RyanEwen:
hmm I just realized this might rely on the HA autocomplete stuff in order to get the calendars and friendly names...

jtenniswood:
Could you break these down into separate PRs. Thanks

RyanEwen:
> Could you break these down into separate PRs. Thanks

Will do (probably closer to the weekend)

jtenniswood:
I think this is a bit complex for the current setup, I will capture both as an idea for future development. Thanks!

PR #1249, opened 2026-07-21
Sunton ESP32-8048S043

RyanEwen:
Adds support for the Sunton ESP32-8048S043, an 800x480 landscape ESP32-S3 panel with an RGB parallel display and GT911 touch. The same definition also covers the 5 inch ESP32-8048S050, which shares the board layout.

No new display model was needed: the existing `ESP32-8048S050` entry in `components/mipi_rgb/models/sunton.py` already matches this board's pin map and porch timings exactly.

Verified on hardware — boots, renders, touch works, web UI loads, and rotation works at 0/90/180/270.

## Three pre-existing fixes this surfaced

Each was latent and invisible until a non-square RGB panel exercised the code. They're separated into their own commits and stand alone if you'd prefer them split out.

**`swap_xy=UNDEFINED` on the Sunton models no longer validates.** `DriverChip.transforms` now raises instead of inferring "no hardware transforms". Because it's a plain `ValueError` rather than a validation error, it surfaced as a bare `not a valid value` naming no key. Unnoticed because the models were unused and every shipping device descends from `ST7701S`, which overrides `transforms`. These panels have no init sequence and therefore no MADCTL register, so an empty `transforms` set is declared explicitly.

**`check_web_smoke.js` validated every device against the wrong feature block.** It matched `features:{...}` across the whole bundle and took the first hit. This passed only because all shipping devices share `displayOffset: 180` (and `esp32-p4-86` declares none, so it's skipped). Any other offset fails against a block that isn't its own.

**The five setup screens broke at 90/270 on non-square panels.** They pinned their root container to `$screen_width x $screen_height`, but LVGL swaps the logical canvas for software rotation, so the container overflowed and pushed content off screen. Square panels are immune because the swap is a no-op — which is why only square devices were ever rotated successfully. `100%` resolves to identical pixel values at 0 degrees; verified by compiling the 4848S040 before and after, with RAM identical to the byte.

## Notes on the profile

- **Grid is 3x2, reflowing to 2x3 rotated**, mirroring the JC4880P443 transposed for landscape so configs move between the panels slot-for-slot.
- **New `wide-landscape-constrained` modal profile** — no existing profile combined a landscape layout family with the constrained memory tier.
- **None of the S3 low-memory build flags are carried over.** With the full card set compiled in, the panel measures 99KB free heap and 3.8MB free PSRAM, so `DISABLE_WEATHER_FORECAST`, `DISABLE_TODO`, `LOW_HEAP_MEDIA_CONTROL`, `LOW_HEAP_COVER_ART` and the reduced image-slot count are all unnecessary here.

## Verification

- `npm run check:product` and all device checks pass (`device-profiles`, `device-slots`, `device-matrix`, `firmware-modal-layouts`, `card-runtime-coverage`, `release-confidence`).
- All seven devices validate; the 4848S040 fully recompiles unchanged.
- One pre-existing `web-smoke` failure ("date preview follows selected language month names") reproduces identically on a clean checkout of `main` and is unrelated to this branch.


jtenniswood:

RyanEwen:
> Sorry, I can't add devices I don't own, as support becomes a nightmare, ending in a poor end user experience. Happy to keep this as a community version for others, as long as it's clear it's officially unsupported

All good. This device is a pain in the ass anyway and I am not sure I want to bother with it anymore, haha. S3 devices just aren't up to the task for the things I am trying to do with them.

jtenniswood:
Yes that's where I ended up too. They're slow, and memory starved. It's why I push people towards the p4 devices, as it's not a lot more for a far better device.

The s3 4inch is endearing because it just keeps pushing on, has a bright screen and is £16! Crazy. The rest of them I can leave behind

PR #1281, opened 2026-07-28
Serve the panel configuration over HTTP at /api/config

landonr:
## What this does

The panel could only be configured from the browser setup page. Card and subpage settings live in internal entities that never reach Home Assistant, so automations could not touch them at all, and there was no headless path for scripts.

This serves the same `espcontrol.backup` v2 envelope over HTTP, for reads and writes:

| Method | Path | Purpose |
| --- | --- | --- |
| `GET` | `/api/config[?section=…]` | The envelope, or one section, with an ETag |
| `POST` | `/api/config/set` | Change one value or a handful |
| `POST` | `/api/config/import` | Apply a whole envelope, atomically by default |
| `POST` | `/api/config/apply` | Commit to flash, optionally reboot |

Cards carry the `raw` string the panel actually stores, so a panel-to-panel copy or a restore is byte-exact rather than re-encoded. Import validates everything first and applies nothing if any value fails, so a rejected value cannot leave the panel half-configured.

User documentation is in `docs/reference/config-api.md`, including Home Assistant `rest_command` and `rest` sensor examples.

## Practical impact

Automations and scripts can now read and change any panel setting, including card and subpage configuration that Home Assistant has never been able to see. Nothing about the existing browser page changes in this PR — it still writes entities directly. Moving it onto this API is a follow-up.

## Security

**This adds no authentication and no new exposure.** The web server already lets anyone on the network read and change every configuration value with no credentials — `curl -X POST 'http://panel/text/Button%201%20Config/set?value=~evil'` has always worked. The new endpoints are that same capability in a more usable shape.

They are registered through `global_web_server_base->add_handler()`, so they inherit `web_server: auth:` if it is ever enabled. That is strictly better than the two existing custom endpoints (`/local_actions`, `/local_sensors`), which called the raw server and bypassed auth entirely — fixed here, along with their `canHandle` prefix matches, which accepted `/local_sensorsXYZ` and could shadow a later handler.

Closing the pre-existing unauthenticated write surface is deliberately out of scope; `docs/reference/config-api.md` documents the posture and the options.

## Design notes

Four things are load-bearing, so each is explained in `dev-docs/firmware.md` **and** guarded by `check_firmware_ha_bindings.py` with negative self-tests:

- **Entity writes must run on the main loop.** A template entity's `set_action:` drives LVGL and the scheduler; writing from the httpd task hard-locks the panel — pingable, no HTTP, no serial, until a power cycle.
- **Responses stream, never materialise.** After an import's validation pass, internal RAM is fragmented to a ~10 KB largest free block with ~28 KB free. A `std::string` growing past ~5 KB gets a null `malloc`, and since `operator new` cannot throw in this build the panel panics and reboots. This was found the hard way.
- **Reading a JSON body depends on fork behaviour.** `web_server_idf.cpp` falls through for unrecognised content types *before* consuming the socket. A fork sync that reorders that would not fail to compile — it would silently make import see an empty body.
- **Registration goes through the web server base**, which is what applies auth.

The envelope-to-entity map is generated from `entity_names.json` and gated in both directions: add a setting and the build fails until it declares a `backup` mapping or a `backupUnmapped` reason; add a field to the export literal and it fails until the firmware can write it. Both directions were verified by breaking them on purpose.

## Testing notes

**Automated:** `check:fast` passes apart from two failures that are present on clean `main` in the same environment — `web-smoke`'s "date preview follows selected language month names" and a stale `calendar: date` fingerprint in `card_runtime_surface_baseline.json`. Both concern the same date preview and reproduce with this branch's registry files reverted to `main`'s, so they are not from this change. I deliberately did **not** regenerate that fixture, since committing this machine's output would break CI.

**Compiled:** ESP32-P4 (`esp32-p4-86`) builds clean — RAM 40.6%, flash 42.1%. The S3 could not be built locally: ESP-IDF's generated mbedtls certificate bundle fails to assemble on clean `main` too, so that target is left to CI.

**Not yet tested on hardware against this base.** The same implementation was exercised end to end on a 4" S3 panel on an earlier base — full byte-exact round trip, atomic abort with a 65-key envelope leaving the panel untouched, `?if_match` returning 412, worst case 24 writes in 0.31 s, and OTA in both directions with config intact. That was before this port, so it needs re-confirming on real hardware before merge. Suggested check on an S3 or P4 panel:

```bash
curl -s http://<panel>/api/config > before.json
curl -s -X POST -H 'Content-Type: application/json' --data-binary @before.json \
  'http://<panel>/api/config/import?dry_run=1'
curl -s -X POST -H 'Content-Type: application/json' --data-binary @before.json \
  http://<panel>/api/config/import
curl -s http://<panel>/api/config | diff - before.json
```


landonr:
Closing to reopen as a single PR — the browser configurator rewire onto this API belongs in the same change, since the endpoint and its only consumer should land together.

PR #1282, opened 2026-07-29
Clock bar night mode icon

landonr:
## Summary

- adds an optional moon to the clock bar that shows while the night schedule is in its night period. off by default, toggle is `Show Night Mode Icon` under Settings > Clock Bar.
- reads the night decision `display_mode_reconcile` already resolves instead of recomputing the schedule, so Time and Sensor mode both work, including the Sensor Is On/Off activation choice.
- first commit is a prerequisite refactor with no feature in it. battery and voice-mute were both hardcoded to the same offset half an item gap left of the network icon, so a moon there would have landed on top of them. that offset is also measured against the tap target rather than the glyph inside it, which is why the spacing looked like a whole missing icon. now theres a cursor that packs icons leftwards by glyph edge, 8px apart.
- new `Screen: Clock Bar Night Mode Icon` switch restores off and is carried in config backups. relayout is change-guarded so the every-minute schedule check doesnt relay the bar out on every tick.

## Practical impact

- you can tell at a glance that the panel is in its night period without reading brightness or guessing.
- most useful with the At Night Time actions that leave the panel usable (Screen Dimmed / Clock). Screen Off hides the whole clock bar anyway so theres nothing to show there.
- existing battery and voice icons move a few px closer to the network icon as a side effect of the packing change. battery is 12.8" only, voice is P4-86 only.
- the moon remains off until its toggle is enabled; independently, the packing refactor adjusts the existing battery and voice icon positions and removes the reserved network gap when connectivity is hidden.

## Documentation decision

- [x] Updated public docs or release-facing notes for user-visible behavior/configuration.

Docs notes:

- `docs/features/clock-bar.md` covers the new toggle and why its only visible with the night actions that leave the panel usable. `docs/features/screen-schedule.md` points at it from the schedule side.

## Testing

- the moon widget is added to all 6 panels. flashed and tested on the **4" S3 (guition-esp32-s3-4848s040)** only.
- **12.8" (jc8012p4a1)** and **P4-86** are worth a look even without using the feature, since the packing change moves their existing battery / voice icons.

## Notes for testing

- turn on Clock Bar plus Show Night Mode Icon, set Night Schedule to Screen Dimmed (not Screen Off, that hides the bar), then cross the night hour. or use Sensor mode and toggle the sensor entity.
- the 12.8" is the only panel that can show two optional icons at once, so worth checking the battery icon and the moon together there.
- all six factory firmware targets compiled with the repository-pinned ESPHome 2026.7.0 environment on the tested packing fix.
- the 10-inch V1 was flashed over OTA with the packing fix and returned online; visual layout confirmation and the P4-86 voice-icon/tap-target test remain pending.
- after rebasing the packing fix onto the latest PR head, `git diff --check`, the firmware parser, generated-device check, and all 34 `check:fast` tasks pass.
- two check failures already exist on `main` and arent from this branch: `check_web_smoke.js` (locale month names) and `check_firmware_ha_bindings.py` (stale local `.esphome/` artifacts, local only).
- no moon in the web preview, same as the battery icon, since both depend on runtime state. can add a ghosted one if thats preferred.
- the icon is `mdi:brightness-3`. `mdi:weather-night` is an outline crescent with outline stars and reads as hollow at 26px.

## PR status

- [x] Needs automated fix.
- [ ] Ready for device test.
- [ ] Ready to merge after user confirmation.

## Issue handling

image


jtenniswood (changes requested):
Merge blocker: the live PR still seeds the optional-icon row from the network glyph unconditionally, so hiding connectivity leaves the network-sized gap that the 10-inch test build was intended to remove.

The exact tested correction is available as commit b4579187d808b488a28eca267326d8a88423fc0d:
https://github.com/jtenniswood/espcontrol/commit/b4579187d808b488a28eca267326d8a88423fc0d

Please cherry-pick that commit onto `clock-bar-night-mode-icon`. It applies cleanly to the current head. Verification on the rebased commit: `git diff --check`, firmware parser tests, generated-device check, and all 34 `npm run check:fast` tasks pass. The same packing change compiled for all six factory targets with ESPHome 2026.7.0 and was flashed successfully to the 10-inch V1.

The P4-86 voice-icon position and both clickable targets still need physical confirmation before merge.

jtenniswood:
I resolved the conflicts against the current `main` and preserved the clock-bar packing fix in draft replacement PR #1303: https://github.com/jtenniswood/espcontrol/pull/1303

The contributor fork branch could not be updated directly, so #1303 is the maintainer-owned, conflict-free continuation. Automated local checks and the three pinned ESPHome 2026.7.0 factory compiles pass. Physical 12.8-inch and P4-86 verification is still required before merging.

jtenniswood:
Thanks for the contribution. This work was carried forward in #1303 because maintainer commits could not be pushed to the contributor fork branch.

#1303 preserved the original clock-bar night mode icon work, included the clock-bar packing correction, and has now been merged into `main`. This PR is therefore superseded and is being closed in favour of the merged replacement.

Issue #1287, opened 2026-07-31
QR Code card support for sharing Wi-Fi credentials


txptr:
### Is your feature request related to a problem? Please describe.
When hosting guests, sharing Wi-Fi credentials (or quick links) often involves manually reading passwords or looking up a static paper printout. Having a quick-access card on the display panel to show a QR code directly would significantly improve user experience for guest access.

### Describe the solution you'd like
I would like to request a new card type (e.g., `qrcode` or `qr`) that generates and displays a QR code on the panel screen. 

**Key Use Cases:**
* **Guest Wi-Fi:** Displaying a `WIFI:S:<SSID>;T:<WPA|WEP>;P:<PASSWORD>;;` formatted string so guests can easily join the network by scanning the display.
* **Quick Links:** Sharing URLs (e.g., local dashboard links, guest control pages).
* **Text / Data:** Sharing plain text or authentication tokens.

### Proposed Configuration Example
Ideally, the card could be configured in YAML similar to other cards:

```yaml
cards:
  - type: qrcode
    title: "Guest Wi-Fi"
    data: "WIFI:S:MyGuestNetwork;T:WPA;P:SecretPassword123;;"
    # Optional parameters:
    size: 160          # Size in pixels
    scale: 4           # QR code scale factor
    icon: "mdi:wifi"   # Optional header icon

roycepope:
Great idea! Here is a mockup:

<img width="436" height="434" alt="Image" src="https://github.com/user-attachments/assets/246d3fe4-ab64-4cbf-9377-381c077a7c8e" />

txptr:
Either like this or with a button showing a dedicated card with just the qr code which would it make easier to scan especially if you have more than one QR code on the layout. What do you think?

jtenniswood:
Good idea. I think a share button, then a modal that shows a QR code. 
Will have to figure out how to generate those locally or remotely.

txptr:
If I'm right lvgl has some kind of library for this https://lvgl.io/docs/open/8.3/libs/qrcode 🥇

talk-more:
Also maybe instead of a specific card for this have it available when clicking the wifi button in the top right?

txptr:
@talk-more Maybe I got you wrong but I don't like the idea. I don't want to share my usual wifi wo everyone having access to my panel.

In my setup I run three separated wifi networks: Internal, iot and guest wifi 

I would like to share guest wifi but not the iot where my panel is connected to.

[...]

PR #1442, opened 2026-08-10
Add Wi-Fi Share QR card

jtenniswood:
## Summary
Adds the **Wi-Fi Share** card requested in #1287.

The card stores the SSID, security type, password, and hidden-network setting in the dashboard configuration, then shows a locally generated QR code when tapped. It is supported on both the main dashboard and subpages. QR payload handling escapes special characters correctly and never writes passwords to device logs.

The editor clearly warns that the configuration is encoding rather than encryption, and the backup area now warns that exports can include Wi-Fi Share passwords.

## Checks
- `npm run check:types`
- `npm run test:web-unit` (24 passed)
- `python3 scripts/check_card_contract_outputs.py`
- `python3 scripts/check_firmware_card_runtime.py`
- `python3 scripts/check_firmware_parser.py`
- Firmware CMake tests (27 passed, including Wi-Fi QR payload cases)
- `npm run docs:build`

A full ESPHome build was started for the JC1060 P4 configuration. It completed configuration generation but timed out while ESP-IDF was still being extracted, before project sources compiled. CI and a device flash remain the final confirmation.

## Device testing
Flash a supported display and verify:
1. WPA/WPA2 and Open networks join from Android and iOS.
2. Hidden-network mode works.
3. SSIDs/passwords with punctuation and non-ASCII characters work.
4. The modal closes cleanly and credentials never appear on the tile or in logs.

Relates to #1287.


jtenniswood:
Replaced with new version

PR #1465, opened 2026-08-11
Migrate sensor Product Model pilot

jtenniswood:
## Summary
- makes the Sensor card and Guition JC8012P4A1 (10-inch V1) device entry authoritative Product Model v2 sources
- routes card and device generation through the Product Model while preserving current generated output
- adds parity checks and documents how subsequent card and device migrations should work

## Validation
- `npm run check:product`
- `npm run check:types`
- `npm run test:web-unit`
- `python3 scripts/build.py --check`

## Device testing
Physical flashing is intentionally deferred to the final integrated refactor milestone. The target for that final pass is the 10-inch V1.


jtenniswood:
Fixed the sparse firmware checkout in 1c92dd2b. Both matrix workflows now include the Product Model inputs explicitly, and `python3 scripts/check_firmware_release.py` passes locally.

PR #1472, opened 2026-08-11
Extract sensor card mode controller

jtenniswood:
## Purpose
Move the sensor card editor’s source and display-mode transitions into a typed controller.

## Practical impact
Selecting Home Assistant/local sources or numeric, time, text, and icon display modes keeps the existing editor behaviour and persistence order, while the reset rules are now independently tested.

## Checked
- `npm run check:types`
- `npm run test:web-unit` (25 passing tests)
- `npm run check:web-smoke`
- `npm run check:product`

## Device testing
No physical flashing for this browser-only change. Once CI is green, test a sensor card on the 10-inch V1 (`guition-esp32-p4-jc8012p4a1`): change its source and each display mode, save, reload, and reconnect.


jtenniswood:
Addressed: removed the three unreferenced intermediate bundles (`67800f…`, `72e375…`, and `b52773…`). The generated-output and web smoke checks pass with only the manifest-selected bundle retained for this PR.

jtenniswood:
Addressed the follow-up: removed the stale `be182170…` bundle. `web-assets.json` continues to select the current bundle, and `npm run check:web-smoke` passes.

PR #1474, opened 2026-08-11
Import feature controllers directly

jtenniswood:
## Purpose
Move the remaining typed feature-controller factories out of the global bootstrap and into the settings and state modules that own them.

## Practical impact
Clock bar, schedule, voice services, screensaver, cover-art, media playback, and alarm-delay settings retain their behaviour while their controllers now use explicit imports.

## Checked
- `npm run check:types`
- `npm run test:web-unit` (24 passing tests)
- `npm run check:web-smoke`
- `npm run check:product`

## Device testing
No physical flashing for this browser-only refactor. After its parent and CI are green, test the 10-inch V1 (`guition-esp32-p4-jc8012p4a1`) settings journey: clock bar, night schedule, voice services, screensaver, cover art, save, reload, and reconnect.


jtenniswood:
Addressed: removed the three unreferenced intermediate bundles (`37b7aa…`, `d5d13d…`, and `df16fc…`). The current manifest-selected bundle remains and `npm run check:web-smoke` passes.

PR #1668, opened 2026-08-15
Add JC1060P470 V2 (new panel) device profile

### LINKED ISSUE #1273: Guition JC1060P470 v2 support
See attached image - I get this visual glitch after installing ESPcontrol. I got it via the browser install and ESPhome install as well. Prior to flashing, it showed a clean image of the default "OS" on the tablet, but I don't know how to flash back to that firmware to see if it still works. Any suggestions? Need logs?

Image

mkucharek:
## Purpose

Guition ships a second 7-inch panel under the same **JC1060P470** name since ~mid-2026 (rear label shows `V2` in the SKU / material number; PCB silkscreen `JC1060P470C_I_W_Y`). Same JD9165 driver IC, but it needs a **different vendor init table and DPI timings**. The bundled ESPHome `JC1060P470` model still sends the old table, so these units show a white screen with a vertical noise band — #1273 (three reporters + my own unit).

This adds `guition-esp32-p4-jc1060p470-v2`, mirroring how the JC8012P4A1 V2 split was done.

## Practical impact

- New install option "JC1060P470 new panel" (installer + C6 recovery selectors label the pair "original panel — no V2 on the rear label" / "new panel — rear label marked V2", like the JC8012P4A1 pair): web installer button, factory/OTA/recovery builds, manual-ESPHome package path, docs on how to identify the panel.
- Original `guition-esp32-p4-jc1060p470` profile is untouched (its generated outputs are byte-identical; the catalog now shares `display`/`artwork` profiles between the two, like `jc8012p4a1` / `-v2`).
- The v2 device dir is identical to v1 except:
  - `display:` block — `model: CUSTOM` + New_Panel init table/timings (h 24/136/160, v 2/21/12, pclk 52 MHz, still 2 lanes @ 750 Mbps RGB565);
  - `esp32_hosted.sdio_frequency: 10MHz` (was 20 MHz) — see below;
  - ids/log tags. The `espcontrol.panel_config` (native configuration document) block is kept identical to v1 — an earlier revision of this branch had dropped it, which silently put V2 units on the legacy text-entity path; fixed in `eef8e259b`.

Source of the table/timings: Guition's reference driver `JC1060P470C_I_W_Y.zip` (packaged 2026-07-13, `pan.jczn1688.com`) → `1-Demo/.../JC1060P470C_I_W_Y_New_Panel/lvgl_demo_v8/src/lcd/esp_lcd_jd9165.{c,h}` (identical in the IDF demos). Guition's burn notes: *"For those with 'V2' after the material number on the label on the back of the shell, please program a new panel."*

## SDIO 20 → 10 MHz (second finding on these boards)

With the correct panel table, these units still showed a rock-steady band of vertical streaks in a fixed column range (~90–250 px from the left) whenever **ESP-Hosted was active** (AP or STA). Bisected with minimal ESPHome images changing one variable each:

- unaffected by DPI porches/pclk (old vs new timings), backlight level, LVGL rotation (band stays put), AP vs STA;
- clean with the same LVGL page and `esp32_hosted` removed;
- DSI lane rate 1000 Mbps breaks the link outright (JD9165 can't decode above 750);
- **`sdio_frequency: 10MHz` removes the band completely** (LVGL test image and the full EspControl image).

So the ESP32-C6 SDIO link at 20 MHz disturbs the DPI framebuffer fetch on these boards (all reporters so far are ESP32-P4 **rev 1.3**); halving the SDIO clock takes it below the threshold. Cost is a lower hosted-link ceiling (~5 MB/s), irrelevant for HA/API traffic and cover art; only C6 firmware OTA gets slower. Only the V2 profile is changed.

## How it was checked

Automated (all pass): `npm run check:device-matrix`, `check:device-profiles`, `check:product` (19 tasks), `check_tasks.py run-task device-manifest` / `device-slots`, generators idempotent, `esphome config` valid for `builds/guition-esp32-p4-jc1060p470-v2.{yaml,factory.yaml}` and `devices/.../dev.yaml`; `check:fast` blocked only by missing local cmake (firmware-tests), everything else in it passed individually.

Physical (my JC1060P470C_I_W_Y, P4 rev 1.3, ESPHome 2026.7.4):
- ESPHome test card + LVGL test pages with this display block: correct colours/geometry, both with these timings and with upstream's old timings.
- Full v2 factory image (this branch): WiFi setup via captive portal, HA pairing, card grid, touch, rotation, cover art all working; no streak band; WiFi/HA connection solid.
- Control: stock v1 image on the same unit → garbled as in #1273. So the table is required.

## Known cosmetic issue (not blocking, help wanted)

On my unit the boot/loading page ("Starting up" / "Connecting to WiFi", black background) does not render: from backlight-on until the first real page (card grid / WiFi-setup page) the panel shows a uniform lavender, finely striped full-screen pattern with no text. Everything after that is clean. Same look as the streak band at 100% line width and as a broken DSI link, so it's a link/fetch disturbance during the boot phase rather than LVGL content; not affected by the SDIO change. Minimal LVGL images boot clean on the same unit. Not yet root-caused — if you have a stock (non-V2) 7" at hand: does it show the boot text? Reports from other V2 owners welcome (with the `ESP32 Chip: ESP32-P4 revX.Y` log line).

## Testing notes

Affected device: 7-inch Guition JC1060P470 with `V2` on the rear label. Flash `guition-esp32-p4-jc1060p470-v2` (web installer button on the 7-inch page from this branch's docs, or ESPHome with `devices/guition-esp32-p4-jc1060p470-v2/packages.yaml`). Units without `V2` must keep the original profile — the wrong one gives a white/noisy screen; reflash the other profile over USB to recover.

Fixes #1273


jtenniswood:
Ah excellent, I couldn't find any fix for the v2, I suspected it was the case, as the 10inch was the same.

jtenniswood:
Could you undo the back of the case and let me know what four digits are on there, as it's useful to guide users by that data as to what version they should be using. Thanks!

jtenniswood:
Because I can't test this, let me know when it's ready to merge.

mkucharek:
> Could you undo the back of the case and let me know what four digits are on there, as it's useful to guide users by that data as to what version they should be using. Thanks!

yep, the digits are `2627`. 

> Because I can't test this, let me know when it's ready to merge.

sure! I did some work today but I'll be out for the rest of the week - I will pick it up over the weekend and hopefully get it ready then!

jtenniswood:
Thanks, just let me know, I know this is coming up frequently so it would be good to get this out soon.

mkucharek:
This is ready to merge from my side @jtenniswood.

## What changed since the last update

- Merged current `main` (through `2c47671d3`). Conflicts were the install selector and three generated web artifacts; the generated ones were regenerated with `scripts/build.py`, `generate_device_manifest.py` and `generate_device_slots.py` rather than hand-merged, and the `web_migration_baseline.json` bundle sizes were recomputed.
- Picked up the ESPHome 2026.8.0 bump and re-verified the panel against it.
- Ran a review pass over the branch and applied the findings worth applying (details at the bottom).

## Hardware verification

Built the factory image and flashed it with a full chip erase to a `JC1060P470C_I_W_Y` / ESP32-P4 rev 1.3 unit, so it came up genuinely un-provisioned.

- First boot shows **"Configure your WiFi / Connect to 'EspControl 7inch V2'" and it stays on screen.** #1720 fixes the disappearing-onboarding behaviour on P4/7-inch too, not just the S3 it was tested on.
- Panel init is healthy on both a warm reboot and a cold power-on: no white screen, no vertical noise band, no garbled output.
- Boot log: ESPHome 2026.8.0, ESP32-P4 rev 103, ESP-IDF v5.5.5, boot-loop counter reset cleanly.

## How this profile differs from the original

Of the 13 files in the profile, **9 are byte-identical** to `guition-esp32-p4-jc1060p470` and 4 differ. Only two of those differences are functional:

| File | Difference |
|---|---|
| `device/device.yaml` | **Display block only.** `model: JC1060P470` → `CUSTOM`, plus the explicit DPI timings and the 58-line JD9165 vendor init table for the new panel. `reset_pin: GPIO05` → `GPIO0`. Added `setup_priority: 850`. Everything outside the `display:` block is unchanged. |
| `device/network_coprocessor.yaml` | `sdio_frequency: 20MHz` → `10MHz` |
| `dev.yaml` | device name suffix only |
| `README.md` | panel-identification text |

Byte-identical: `.gitignore`, `esphome.yaml`, `packages.yaml`, `device/fonts.yaml`, `device/lvgl.yaml`, `device/sensors.yaml`, `device/network_coprocessor.yaml`'s Ethernet sibling, `device/starter_network.yaml`, `device/starter_network_ethernet.yaml`.

### On the two functional changes

**`reset_pin: GPIO0`** — the `JC1060P470C_I_W_Y-V1.0` schematic has FPC1 pin 4 (`LCD_RST`) on GPIO0; GPIO5 only reaches the expansion header, so both existing 7-inch profiles never actually reset the panel, and configuring a reset pin also suppresses ESPHome's software reset. Confirmed on hardware by driving GPIO0 low from the web UI, which blanked the panel. With a real reset pin the 120 ms settle delays in `init_sequence` become mandatory — without them the panel comes up white on a warm reboot. This is worth a look for the original profile too, but I have not changed it there since I can't test that panel.

**`sdio_frequency: 10MHz`** — at 20 MHz these boards show a fixed band of vertical streaks in the MIPI-DSI output, independent of DPI timings, lane rate, backlight and rotation. 10 MHz removes it completely. **Caveat I want to flag rather than bury:** the original profile's comment justifies 20 MHz with "camera transfers remain supported at this speed", and I have not measured camera throughput at 10 MHz. Home Assistant traffic and cover art are fine. If you'd rather not halve the link budget for camera users, the alternative is to treat the streak band as a known cosmetic issue on this revision — your call, and I've written the caveat into the YAML comment either way.

## On the case digits

My new-panel unit reads **2627** (PCB silkscreen `JC1060P470C_I`). Worth knowing before you use that to guide users: on the 10-inch, 2627 means *original* panel, so the threshold doesn't carry over to the 7-inch, and a single data point can't establish one. The rear label's `V2` marking in the SKU / material number is the reliable discriminator so far, and that's what the install selector and docs key off. Collecting digits from the reporters in #1273 would be the way to establish a 7-inch threshold if you want one.

## Known cosmetic residual

A sub-second "lavender checkerboard" flash right after display-on on a cold boot, which then snaps to true black. It's the JD9165's no-valid-video colour showing in the gap between `DISPON` and LVGL's first flush, with the backlight already lit. Closing it properly means holding the backlight off until the first flush, which is a change to shared backlight behaviour rather than something I'd put in a device profile.

mkucharek:
@jtenniswood let me know if there's anything I can do here!

jtenniswood:
Awesome, many thanks for all the work!

mkucharek:
my pleasure, just wanted to say that this is an awesome project @jtenniswood  and I'm happy to be able to contribute!

PR #1675, opened 2026-08-16
Fix camera tile label clipping and misaligned drop shadow

### LINKED ISSUE #1674: Camera (image) card tile labels clip glyphs and sit at the wrong height
Two rendering problems with the tile label on image (camera) cards, on the home grid. Confirmed on a Guition ESP32-P4 JC1060P470 (7 inch), present on current main.

**1. Label text is clipped at the bottom**
The bottom of the glyphs is cut off - descenders like the loop of a "g" (e.g. "Package"), and the bottom edge of other letters too.

Cause: image cards position the tile label with a bespoke helper that sets the label to the button's full width and copies the button's 16px padding onto the label. With the label's fixed height, that squeezes the usable text area below the font's line height, so `LV_LABEL_LONG_WRAP` clips the bottom of the text inside the label itself. (Moving the label up does not help - the clip is the label's own box, not the tile edge.)

Every other card lays its label out the same simple way (`LV_LABEL_LONG_WRAP`, a percentage width that already respects the button padding for the side margin, natural content height, bottom-left align). Image cards should do the same.

**2. Drop shadow is misaligned**
The label is drawn twice (a dark copy behind a white copy, for legibility over the image). The two copies are positioned independently, so the dark "shadow" drifts out of alignment - above and to the left of the text - instead of sitting just behind it.

I have a fix prepared and will open a PR shortly.

jsenecal:
## Summary

Fixes #1674. Makes image (camera) card tile labels use the same layout as every other card so the text stops clipping, and pins the label's drop shadow so it no longer drifts.

- Clipping: image cards positioned the tile label with a bespoke helper that forced the label to the button's full width and copied the button's 16px padding onto the label. With the label's fixed height, the usable text area dropped below the font line height, so `LV_LABEL_LONG_WRAP` clipped the bottom of the glyphs inside the label. Replace it with the standard card-label layout: `LV_LABEL_LONG_WRAP`, a percentage width (which already respects the button padding for the side margin), the label's natural content height, and a bottom-left align. As a side effect the label now lines up with the standard card-label height.
- Drop shadow: the label is drawn twice (a dark copy behind a white copy for legibility over the image). Each copy was aligned independently, and LVGL resolves a bottom-left align per object, so the two drifted apart. Give the shadow the same layout as the label and offset the copy by one pixel with a style translate, which LVGL re-applies on every layout pass, so the shadow stays locked one pixel below-right of the text.

## Practical impact

- Camera tile labels render without clipped descenders and keep a tight one-pixel drop shadow. No configuration or saved-data changes.

## Documentation decision

- [ ] Updated public docs or release-facing notes for user-visible behavior/configuration.
- [x] No docs needed because this does not change user-visible behavior/configuration.

Docs notes:

- Rendering fix only; no documented behavior changes.

## Testing

- Guition ESP32-P4 JC1060P470 (7 inch). Built and flashed over OTA.

## Notes for testing

- Diagnosed on device by logging the resolved LVGL geometry (button height 190, label box bottom at the tile edge, font line height 26 vs a 20px usable text area), which showed the clip came from the label's own box rather than the tile edge.
- Firmware host tests pass (26/26).
- Verified on a JC1060P470: camera tile labels ("Door", "Package") no longer clip, align with the other card labels, and the shadow is a clean one-pixel drop shadow.

## PR status

- [ ] Needs automated fix.
- [ ] Ready for device test.
- [x] Ready to merge after user confirmation.

## Issue handling


jtenniswood:
Awesome, thanks, that's great

PR #1847, opened 2026-09-03
Fix unavailable Companion media cards

jtenniswood:
## Purpose
Fix Companion media-control cards being marked unavailable when the macOS MediaRemote command API is present but an optional Now Playing read symbol is missing or changed.

## Change
The MediaRemote bridge now loads the framework independently. Media command availability checks only the command symbol, while Now Playing checks continue to require the read symbols they use.

## Checks
- `swift build` passed.
- Focused C++ Companion protocol test passed.
- Browserless application contract tests passed: 108 tests.
- Full Swift XCTest target could not run on this host because the command-line toolchain has no XCTest module.
- CMake firmware test runner is not installed in this environment.

## Physical testing
Not yet performed. After the matching Companion app and firmware are available, verify Play/Pause, Previous Track, and Next Track on the 4-inch S3 test display.


jtenniswood:
Moved into #1807 in commit ac461b4e1 so the fix is consolidated with the media-controls change.

PR #1848, opened 2026-09-03
Move connection settings into Status tab

jtenniswood:
## Summary
- Move the existing connection address, pairing, and forget controls into the Status tab.
- Keep the settings in a collapsed disclosure panel by default.
- Remove the separate Connection tab to keep status and setup together.

## Checks
- `swift build` passes in `macos/Companion`.
- `swift test` compiles the Companion target but cannot run in this environment because the available Command Line Tools do not provide XCTest.

## Device/app testing
- Run the macOS Companion app from this branch.
- Confirm the Status tab shows connection state and that Connection settings is collapsed initially.
- Expand it and confirm address editing, pairing, opening the device webserver, and forgetting the panel still work.


jtenniswood:
Moved this fix into PR #1432 as commit 5b62bc5a2. Closing the standalone PR.

PR #1852, opened 2026-09-03
Mac Connector: Highlight focused Companion folders

jtenniswood:
## Purpose

Make an Open folder card light up only while its exact approved folder is shown in Finder's front window.

## What changed

- Companion polls Finder focus while connected and publishes the matching opaque folder action ID.
- The firmware applies the existing checked-card visual to folder actions and clears it for another folder, an unrelated Finder location, or another app.
- Added folder matching regression coverage and documented the Finder automation permission.

## Checks

- Swift Companion build passed.
- Standalone firmware Companion regression test passed with warnings treated as errors.
- Full Swift test suite could not run because the installed Command Line Tools do not provide XCTest.
- Physical testing still needed on the 4848S040 with the matching Companion build: open two approved folders in Finder and confirm only the front folder card is lit.


jtenniswood:
Superseded by consolidated PR #1855.

PR #1862, opened 2026-09-04
Make Companion Now Playing and Stats always on

jtenniswood:
## What changed

Now Playing and Mac system statistics start automatically whenever the Companion is connected. The old saved enable/disable preferences are no longer read, so existing installations become enabled after updating.

The Now Playing and System Stats pages remain available as read-only status/diagnostic views, but their enable/disable controls have been removed.

## Permissions

No additional macOS permissions are required. The existing Accessibility permission remains optional and is only used for keyboard-shortcut cards.

## Checks

- `swift build --package-path macos/Companion` passes.
- `git diff --check` passes.
- Confirmed the old sharing preference/toggle references are absent.
- `swift test --package-path macos/Companion` is blocked on this host because the installed Command Line Tools do not provide XCTest (`no such module 'XCTest'`).

## Device test

After the combined Companion PR is available on the test display, pair the matching Companion build and confirm Now Playing and system statistic cards update without enabling any setting. Physical behavior is not verified by the local build.


jtenniswood:
Superseded by consolidated PR #1855; the always-on Now Playing and Stats changes are now committed there.

PR #1867, opened 2026-09-05
Remove duplicate panel IP display

jtenniswood:
## Purpose
Removes the redundant Wi-Fi and Ethernet `IP Address` diagnostic sensors. The ESPHome web page already renders the active panel address in its built-in Connection section, so the extra sensors caused the IP to appear twice.

## Practical impact
The Connection section shows the panel IP once. The on-screen network status continues to obtain the address directly from the active network interface.

## Checks
- `git diff --check`
- `npm run check:config` (unit tests passed; bundle validation was blocked by a missing local VitePress font: `inter-roman-latin.woff2`)

Physical device testing is recommended on the affected display after flashing.

### DIFF
--- a/common/addon/network.yaml
+++ b/common/addon/network.yaml
@@ -69,7 +69,7 @@ sensor:
         - script.execute: network_status_refresh
 
 # ---------------------------------------------------------------------------
-# IP address (text_sensor)
+# Network transport (text_sensor)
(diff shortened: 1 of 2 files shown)


jtenniswood:
Superseded by PR #1855; the functional changes were applied there.

PR #1928, opened 2026-09-11
Refresh web config UI design

justnateg:
## Purpose
Refresh the web config UI so it no longer looks like a default AI-generated dashboard. The current design uses Inter, an indigo accent, pill-shaped buttons everywhere, near-black shadows, and a narrow 960px column.

## Changes (all CSS/token-level, no structural or dependency changes)
- **Typeface**: drop the Inter webfont stack for the native system font stack.
- **Palette**: warm layered dark surfaces + a warm amber/copper accent replacing indigo; hard-coded indigo tints (clockbar selection, add-pill hover, info panel) tokenized to accent variables; slightly warm-toned shadows.
- **Radius**: measured radius scale — compact 10px for buttons/inputs, 12px for cards/modals; pills (9999px) only for badges/chips/toasts/sliders.
- **Layout**: container widened 960px → 1080px; tighter vertical rhythm in settings fields and section headers.
- **Type scale**: explicit tokens (--text-xs … --text-xl) applied consistently to labels, inputs, buttons, section titles.
- **Interactivity**: unified hover/press transitions, focused focus-visible rings on action/selection/apply buttons, polished empty and loading states.
- Regenerated embedded www.js bundle assets.

## How it was checked
- `tsc --noEmit` passes
- `check:web-unit` passes
- `check:web-asset-manifest` passes (after regenerating www assets)
- `check:fast` — passes except for the pre-existing firmware cover_art_activation_test failure caused by a missing Python yaml module in this dev environment (unrelated to this change; fails the same way on main).

## What to test
- Open the web config UI (embedded webserver) on a device or via `python3 scripts/build.py www` + dev server.
- Verify: new system font rendering, warm amber accent + hover states, compact button/input radius (pills only on badges/chips), wider single-column layout, consistent section header spacing, keyboard focus rings visible, empty state styling.
- No physical device flashing required; this only affects the web config UI styling.


jtenniswood:
Thanks for that, I've taken a few of those updates and changed a few, they're in #1931

PR #1984, opened 2026-09-16
Add Notification card type

zacs:
## Summary

- Adds a **Notification** card: shows one text message from a Home Assistant entity and colours the card from a severity attribute on the same entity.
- `information` keeps the normal card colours, `warning` uses a muted yellow with dark text (~8.4:1), `alert` a red with white text (~6.5:1). Common alternatives are recognised (`warn`/`caution`/`medium`; `critical`/`error`/`danger`/`severe`/`high`), so ordinary template sensors work without extra configuration. Anything unrecognised reads as information.
- Laid out like the Date & Time cards: the message sits where those cards put the time, the level names itself on the label line below, where they put the date. Nothing is vertically centred.
- The message marquees when it is wider than the card, at a constant 90 px/sec. Tapping opens the full text in a modal that also names the level in words, so urgency never rests on colour alone. An optional Home Assistant action adds an **Acknowledge** button to that modal.

CC30267A-E995-4049-8140-8E064D3C561F_1_102_a

## Practical impact

- New card type; nothing existing changes behaviour. Cards that are not Notification are untouched.
- No new icons and no new font sizes: the card adds nothing to any device's icon font, and reuses the existing media-title font role, which is defined on all seven panels.
- Settings are the entity, an optional Name, and three optional values: which attribute carries the level (defaults to `notification_level`), an attribute to read the message from when it exceeds Home Assistant's 255-character state limit, and a Home Assistant action for the Acknowledge button.
- `notification` is added to the subpage options allow-list in `normalize_subpage_btn`, the same list Wi-Fi Sharing was added to in #1948. Without it a Notification card on a subpage loses its level, message and acknowledge settings.

## Documentation decision

- [x] Updated public docs or release-facing notes for user-visible behavior/configuration.

Docs notes:

- New `docs/card-types/notifications.md`, wired into the card-types index and VitePress nav. It shows how to build the sensor rather than assuming one: a plain template sensor, a trigger-based sensor driven by state changes, and one driven by events with an automation that posts to it.
- A card shows one message because it shows one line, so the page also gives a template that joins the notices that currently apply and takes its colour from the most urgent, and says when separate cards are the better answer.

## Testing

- Hardware tested only on a **Seeed reTerminal D1001** (8" 800x1280, five-column 5x4 grid, 20 cards), which is a community device from `lamiskin/espcontrol-community-devices`, not one of the seven panels in this repo.
- **None of the seven supported panels have run this card.** The two worth a look are `guition-esp32-p4-jc4880p443` (two-column grid) and `guition-esp32-s3-4848s040` (S3, smallest `font_text_title` at 34px), where the message and label lines have the least room.

## Notes for testing

- `npm run check:ci` passes in full. The browser smoke was also run separately against a local Chromium and passes.
- What was verified on the D1001: card renders at 1x5, colours and level word change with the attribute, the marquee scrolls and is readable, the modal opens with the full text.
- What was **not** verified anywhere: the card on a subpage, the Acknowledge button actually firing its action, and the card at 1x1 and 1x2 sizes. The 1x1 case is the one I would look at first — the message uses the title font with the level on a second line, and the Date & Time cards solve the same squeeze by hiding their label on wide layouts (`apply_wide_large_date_time_card_layout`). If the level word clips on a small card, that is the pattern to copy.
- Two details found the hard way on device, in case they are useful elsewhere: the value font used by the Date & Time cards is subset to digits, so it cannot render a sentence; and LVGL's speed-based label scrolling clamps to 10s, which makes the effective speed rise with message length, so the card sets the scroll duration from the measured text width instead.

## PR status

- [ ] Needs automated fix.
- [ ] Ready for device test.
- [x] Ready to merge after user confirmation.

## Issue handling

---

https://claude.ai/code/session_017fUHVVXFco6duYiHqhW6Hz


zacs:
A bit more human TLDR in case it's helpful: 

New things it does that you may have an opinion on:

- Adds colors to cards aside from the one configurable highlight color. (FWIW I would like to make the warning/alert colors configurable or snapped to the primary color, but wanted to keep this PR reasonable in size.)
- Adds marquee behavior on a card. 
- Adds the notion of an espControl tap firing an event back on HA. (At least I think this is new.) 

Things that I very purposefully did to keep in line with the repo’s existing design decisions: 
- No new fonts: used the sizes and glyphs from existing cards. 
- No new positions: used the existing layour from the datetime card. 
- No new modal: used the existing one and its patterns.

Issue #50, opened 2026-04-18
Buttons in sub pages don't work correctl


games2day:
Buttons in sub pages don't work correctly. When pressed it will always sent the most bottom button label

<img width="549" height="541" alt="Image" src="https://github.com/user-attachments/assets/e33547ad-1be0-451b-80e3-f19def1f4454" />
<img width="389" height="233" alt="Image" src="https://github.com/user-attachments/assets/ccaf3458-a2d9-49fd-bd52-0f88f0539971" />

jtenniswood:
Should be fixed, try it with the new release.
Either build from main or wait for the firmware to compile and update

PR #1994, opened 2026-09-17
Add LilyGO T-Display S3 Long device profile

willumpie82:
## Summary

Adds a device profile for the [LilyGO T-Display S3 Long](https://lilygo.cc/products/t-display-s3-long) — a 3.4" 640x180 QSPI bar display (AXS15231B panel + CST3530 capacitive touch), landscape orientation, 4x1 tile row.

- Uses ESPHome's stock `mipi_spi`/`AXS15231` driver at the panel's true native geometry (180x640, portrait), with rotation handled entirely by LVGL's own software rotation — this chip has no working hardware axis-swap bit (MADCTL_MV), confirmed against ESPHome's source and this panel's known behavior.
- `full_refresh: true` + a full-size LVGL buffer works around a real controller-level quirk on this panel that corrupts partial/odd-coordinate address windows. This is a documented, still-open issue upstream ([Xinyuan-LilyGO/T-Display-S3-Long#44](https://github.com/Xinyuan-LilyGO/T-Display-S3-Long/issues/44)) — I also opened [Xinyuan-LilyGO/T-Display-S3-Long#49](https://github.com/Xinyuan-LilyGO/T-Display-S3-Long/pull/49) with a minimal working example and a hardware-confirmation photo, and [#50](https://github.com/Xinyuan-LilyGO/T-Display-S3-Long/issues/50) with the root-cause write-up.
- No custom/external display component needed.

Also fixes two things the new device profile's manifest entry exposed:

- `scripts/check_firmware_ha_bindings.py`'s camera-refresh-action check hardcoded a single S3 device slug (`guition-esp32-s3-4848s040`) instead of reading each device's actual `imageSlots` capability from the manifest — this device also has `imageSlots: 1` and would have needed the same special-case forever.
- `scripts/check_firmware_modal_layouts.py` gains two fixture-driven exemption flags (`fiveTabRowOverflowsPanel`, `groupedSpeakerModalBelowMinimum`) for panels too narrow/short for the existing tab-row and grouped-speaker-modal geometry assertions to hold as originally written — this panel's 180px short side triggers the narrow-panel case. Both are scoped to the specific fixture entries that need them, with TODOs for the real layout fix (wrapping/scrolling tabs, a compact grouped-speaker layout) rather than a blanket relaxation.

## Test plan

- [x] Flashed to real hardware — clean landscape image, working touch, no corruption
- [x] `python3 scripts/build.py --check` — all generated outputs in sync
- [x] `npm run check:fast` (product, web, docs domains) — all passing
- [x] Two pre-existing, unrelated local failures confirmed via `git stash` against clean `main`: `cover_art_activation_test` (stricter local Apple Clang than whatever CI pins) and `check_local_artifacts.py --self-test` (environment-coupled, reproduces on clean main) — not introduced by this change


jtenniswood:
Sorry, I can't add devices I don't own, as support becomes a nightmare, ending in a poor end user experience. Happy to keep this as a community version for others, as long as it's clear it's officially unsupported.

PR #2004, opened 2026-09-18
Adjust compact camera metadata placement

jtenniswood:
## Purpose
Improve the optional photo metadata position on smaller camera screensaver layouts.

## User impact
- Moves metadata farther from the bottom and left edges on compact panels.
- Reduces the gap between metadata and the clock above it from 8px to 4px.
- Leaves wide-panel placement unchanged.

## Checks
- `python3 scripts/check_photo_metadata.py`
- `git diff --check`

## Device testing
- Flash and check the camera screensaver with metadata enabled on the affected smaller display(s). Physical testing is still required.


jtenniswood:
Transferred these camera screensaver metadata and clock layout updates into PR #1814.

